PR #20, opened 2019-08-14
Feature/refactor

HuiJun:


core/core.gradle
```diff
@@ -2,6 +2,8 @@ apply plugin: 'io.spring.convention.spring-module'
 
 dependencies {
     compile project(':data')
+    compile project(':json')
+
```
dlamoris, line 6: clean up spring dependencies too

crud/src/main/java/org/openmbee/sdvc/crud/controllers/elements/ElementsController.java
```diff
@@ -32,7 +31,7 @@
         @RequestParam(required = false) Map<String, String> params) {
 
         NodeService nodeService = getNodeService(projectId);
-        ElementsResponse res = nodeService.read(projectId, refId, elementId, params);
+        ElementsResponse res = (ElementsResponse) nodeService.read(projectId, refId, elementId, params);
```
dlamoris: can take out the cast

jupyter/src/main/java/org/openmbee/sdvc/jupyter/services/JupyterNodeService.java
```diff
@@ -5,34 +5,40 @@
...
 import org.openmbee.sdvc.jupyter.controllers.NotebooksRequest;
 import org.openmbee.sdvc.jupyter.controllers.NotebooksResponse;
+import org.openmbee.sdvc.rdb.repositories.node.NodeDAO;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.util.Pair;
 import org.springframework.stereotype.Service;
 
 @Service("jupyterNodeService")
-public class JupyterNodeService extends DefaultNodeService implements NodeService {
+public class JupyterNodeService implements NodeService {
```
dlamoris: should revert to before, this depends on logic in default service, and the notebooks functions were specifically for assembling/disassembling notebooks and shouldn't be overriding the default read

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DbContextHolder.java
```diff
@@ -1,4 +1,4 @@
-package org.openmbee.sdvc.crud.config;
+package org.openmbee.sdvc.rdb.config;
```
dlamoris: think this should be in core, since elastic impl also depends on it to look up current project, unless we decide to not use a context object and just pass project and ref everywhere...
it should really just be called ContextHolder...

rdb/src/main/java/org/openmbee/sdvc/rdb/config/ContextObject.java
```diff
@@ -1,4 +1,4 @@
-package org.openmbee.sdvc.crud.config;
+package org.openmbee.sdvc.rdb.config;
```
dlamoris: see comment for DbContextHolder
guess some of these should be in the impl instead, but it's primarily just keeping track of current project and ref

dlamoris (changes requested):
more things to move:
DatabaseDefinitionService from crud to rdb
ExceptionHandlerConfig from core to crud
PersistanceJPAConfig from core to rdb
repositories from core to rdb

dlamoris (approved):

PR #22, opened 2019-08-22
Feature/mysql

dlamoris:


data/src/main/java/org/openmbee/sdvc/data/domains/Base.java
```diff
@@ -20,7 +20,7 @@
     private static final long serialVersionUID = 8389104517465359723L;
 
     @Id
-    @GeneratedValue(strategy = GenerationType.AUTO)
+    @GeneratedValue(strategy = GenerationType.IDENTITY) //use of AUTO with mysql results in TABLE, for postgres auto is fine
```
HuiJun, line 23: Was this also tested with postgres? I thought there was an issue with using identity.

example/src/main/resources/application.properties.example
```diff
@@ -20,13 +20,19 @@ ldap.group.search.filter=uniqueMember={0}
 
 # See core module for example configuration
 spring.datasource.url=jdbc:postgresql://localhost:5432
+#spring.datasource.url=jdbc:mysql://localhost:3306
 spring.datasource.database=mms
 spring.datasource.username=mmsuser
+#spring.datasource.username=root #mysql
```
HuiJun: You can create users in mysql as well, no need to use root.

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/commit/CommitDAOImpl.java
```diff
@@ -90,7 +90,7 @@ public PreparedStatement createPreparedStatement(Connection connection)
             currentExtraCol++;
         }
         if (timestamp != null) {
-            query.append(" AND date_trunc('milliseconds', timestamp) <= ?");
+            query.append(" AND timestamp <= ?");
```
HuiJun, line 99: The date_trunc function was added to the query to resolve an issue with time comparisons in postgres. IIRC, it had to do with timestamp being stored in microseconds. Maybe we need to add a conditional here...

example/src/main/resources/application.properties.example
dlamoris: got lazy with using the mysql docker, we can take it out, it's just for convenience so i can quickly switch comments when testing different db

data/src/main/java/org/openmbee/sdvc/data/domains/Base.java
dlamoris: testing with postgres works, i think it's just less efficient, would be nice if we can use IDENTITY just for mysql

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/commit/CommitDAOImpl.java
dlamoris: i think the only time this would cause a problem is when we take a timestring from the user (the only place is a maxTimestamp query on commits, and it'll have to be exact), otherwise the tests for getting branch history and element at commit works in both. there isn't an equivalent function in mysql, so maybe we can just make a note of it and try to add it back if problems occur

dlamoris: cleaned up timestamps

PR #26, opened 2019-09-23
Feature/perm bootstrap

HuiJun:


crud/crud.gradle
```diff
@@ -3,4 +3,5 @@ apply plugin: 'io.spring.convention.spring-module'
 dependencies {
     compile project(':core')
     compile project(':rdb')
+    compile project(':permissions')
```
dlamoris: i'll change this to be in example, since crud can already use the permission interface from core, it shouldn't depend on a specific impl

dlamoris (approved):

PR #43, opened 2019-11-08
Feature/webhooks

HuiJun:


webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,208 @@
...
+    @PreAuthorize("#projectId == null || @mss.hasProjectPrivilege(authentication, #projectId, 'PROJECT_READ', true)")
+    public ResponseEntity<? extends BaseResponse> handleGet(@PathVariable(required = false) String projectId, Authentication auth) {
+
+        WebhookResponse response = new WebhookResponse();
+        if (projectId != null) {
+            List<Webhook> webhooks = webhookRepository.findAllByProject_ProjectId(projectId);
+            if (webhooks.isEmpty()) {
+                throw new NotFoundException(response.addMessage("No webhooks not found"));
+            }
+            WebhookJson webhookJson = new WebhookJson();
```
dlamoris, line 54: this and last line should be inside for loop

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,208 @@
...
+    public void setMss(MethodSecurityService mss) {
+        this.mss = mss;
+    }
+
+    @Autowired
+    public void setObjectMapper(ObjectMapper om) {
+        this.om = om;
+    }
+
+    @GetMapping(value = {"", "/{projectId}"})
```
dlamoris: think it makes more sense to have /project/{projectId}/webhooks...
and reserve /webhooks for admin/system level hooks that applies across all projects

webhooks/src/main/java/org/openmbee/sdvc/webhooks/config/EventConfig.java
```diff
@@ -0,0 +1,17 @@
+package org.openmbee.sdvc.webhooks.config;
```
dlamoris, line 1: think this should be in core

webhooks/src/main/java/org/openmbee/sdvc/webhooks/components/EventPublisher.java
```diff
@@ -0,0 +1,24 @@
+package org.openmbee.sdvc.webhooks.components;
```
dlamoris: should be in core

crud/src/main/java/org/openmbee/sdvc/crud/services/DefaultNodeService.java
```diff
@@ -81,12 +82,22 @@ public void setNodeGetHelper(NodeGetHelper nodeGetHelper) {
...
+    @Autowired
+    public void setEventPublisher(Optional<EventPublisher> eventPublisher) {
+        this.eventPublisher = eventPublisher;
+    }
+
     @Override
     public ElementsResponse read(String projectId, String refId, String id,
         Map<String, String> params) {
 
+        EventObject event = EventObject.create(projectId, refId, "READ", params);
```
dlamoris: remove

core/src/main/java/org/openmbee/sdvc/core/services/EventService.java
```diff
@@ -0,0 +1,7 @@
+package org.openmbee.sdvc.core.services;
```
dlamoris, line 1: not sure if we need an event service or multiple impls of event publisher, i think just have a publisher ready in core should be enough..

crud/crud.gradle
```diff
@@ -2,4 +2,5 @@ apply plugin: 'io.spring.convention.spring-module'
 
 dependencies {
     compile project(':core')
+    compile project(':webhooks')
```
dlamoris: crud shouldn't depend on webhooks

webhooks/src/main/java/org/openmbee/sdvc/webhooks/services/WebhookEventService.java
```diff
@@ -0,0 +1,23 @@
+package org.openmbee.sdvc.webhooks.services;
```
dlamoris, line 1: not needed?

dlamoris (changes requested):

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
dlamoris: you can keep the /webhooks for now, but also add in projects/projectId/webhooks, the /webhooks behavior will change in the future
the projectid in the url should be used over what's defined in the request for post or deletes

crud/src/main/java/org/openmbee/sdvc/crud/services/DefaultNodeService.java
dlamoris: also needs to add event publishing for commit, ref creation and project creation, i think we can create subclasses of eventobject for CommitEvent, ProjectEvent, BranchEvent so listeners have more flexibility on which to listen to

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+        return ResponseEntity.ok(response);
+    }
+
+    @DeleteMapping
+    @Transactional
+    @PreAuthorize("isAuthenticated()")
+    public ResponseEntity<? extends BaseResponse> handleDelete(
+        @RequestBody WebhookRequest webhookRequest,
+        @RequestParam(required = false, defaultValue = "false") boolean hard) {
```
dlamoris: param not needed

webhooks/src/main/java/org/openmbee/sdvc/webhooks/components/EventListener.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private WebhookDAO eventRepository;
+
+    @Autowired
+    public void setEventRepository(WebhookDAO eventRepository) {
+        this.eventRepository = eventRepository;
+    }
+
+    @Override
+    public void onApplicationEvent(EventObject eventObject) {
```
dlamoris, line 38: for now to support system level webhook, add a env var that can specify a hook and post all events to it if it's configured

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,208 @@
...
+
+        Set<String> projects = new HashSet<>();
+        Set<String> uris = new HashSet<>();
+        for (WebhookJson webhookJson : webhookRequest.getWebhooks()) {
+            projects.add(webhookJson.getProjectId());
+            uris.add(webhookJson.getUri());
+        }
+
+        WebhookResponse response = new WebhookResponse();
+        for (String projectId : projects) {
```
dlamoris: the relationship between webhook and project id is lost, this can inadvertently delete the wrong webhook

core/src/main/java/org/openmbee/sdvc/core/components/EventPublisher.java
```diff
@@ -0,0 +1,24 @@
...
+    @Autowired
+    public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
+        this.applicationEventPublisher = applicationEventPublisher;
+    }
+
+    public void publish(final EventObject event) {
+        applicationEventPublisher.publishEvent(event);
+    }
+
+    public void publish(final String event, final Object payload) {
```
dlamoris: this doesn't actually publish the payload, don't think it's needed

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.security.core.Authentication;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.*;
+
+@RestController
+@RequestMapping("/project")
```
dlamoris: should be /projects
well it could just be /projects/{projectId}/webhooks since they're all the same

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+    public void setMss(MethodSecurityService mss) {
+        this.mss = mss;
+    }
+
+    @Autowired
+    public void setObjectMapper(ObjectMapper om) {
+        this.om = om;
+    }
+
+    @GetMapping(value = {"", "/{projectId}/webhooks"})
```
dlamoris: no need for "" since that'll be projects,
also preauth can take out check for null

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        for (Webhook webhook : webhooks) {
+            WebhookJson webhookJson = new WebhookJson();
+            webhookJson.merge(convertToMap(webhook));
+            response.getWebhooks().add(webhookJson);
+        }
+        return ResponseEntity.ok(response);
+    }
+
+    @PostMapping(value = {"", "/{projectId}/webhooks"})
```
dlamoris: take out ""

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+    @Transactional
+    @PreAuthorize("isAuthenticated()")
```
dlamoris: can just directly check perm with the project

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+    public ResponseEntity<? extends BaseResponse> handlePost( @PathVariable String projectId, @RequestBody WebhookRequest webhooksPost, Authentication auth) {
+
+        if (webhooksPost.getWebhooks().isEmpty()) {
+            throw new BadRequestException(new WebhookResponse().addMessage("No projects provided"));
```
dlamoris: all these messages haven't been updated to reflect webhooks

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        WebhookResponse response = new WebhookResponse();
+        List<Map> rejected = new ArrayList<>();
+        response.setRejected(rejected);
+        for (WebhookJson json: webhooksPost.getWebhooks()) {
+            if (projectId.isEmpty()) {
```
dlamoris: no need to check this since the perm check should already take care of it

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+                rejection.put("code", 403);
+                rejection.put("request", json);
+                rejected.add(rejection);
+                continue;
+            }
+        }
+        return ResponseEntity.ok(response);
+    }
+
+    @DeleteMapping(value = {"", "/{projectId}/webhooks"})
```
dlamoris: remove ""

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+
+        return ResponseEntity.ok(response);
+    }
+
+    public Map<String, Object> convertToMap(Object obj) {
+        return om.convertValue(obj, new TypeReference<Map<String, Object>>() {});
+    }
+
+    private boolean webhookExists(WebhookJson json, String projectId) {
+        List<Webhook> projectWebhooks = webhookRepository.findAllByProject_ProjectId(projectId);
```
dlamoris: why not just make a query that checks for it directly? doesn't have to go through loop each time

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+                rejection.put("message", "Project id missing");
+                rejection.put("code", 400);
+                rejection.put("request", json);
+                rejected.add(rejection);
+                continue;
+            }
+
+            if (!webhookExists(json, projectId)) {
+                try {
+                    if (!mss.hasProjectPrivilege(auth, projectId, Privileges.PROJECT_CREATE_WEBHOOKS.name(), false)) {
```
dlamoris: can just be checked via preauthorize

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+                        Map<String, Object> rejection = new HashMap<>();
+                        rejection.put("message", "No permission to create project under org");
+                        rejection.put("code", 403);
+                        rejection.put("request", json);
+                        rejected.add(rejection);
+                        continue;
+                    }
+
+                    Optional<Project> project = projectRepository.findByProjectId(projectId);
```
dlamoris: can be outside forloop

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
```diff
@@ -0,0 +1,200 @@
...
+                rejected.add(rejection);
+                continue;
+            }
+        }
+        return ResponseEntity.ok(response);
+    }
+
+    @DeleteMapping(value = {"", "/{projectId}/webhooks"})
+    @Transactional
+    @PreAuthorize("isAuthenticated()")
```
dlamoris: same as post

dlamoris (changes requested):
think there's some more comments from before you missed

webhooks/src/main/java/org/openmbee/sdvc/webhooks/controllers/WebhooksController.java
HuiJun: It's because the only other identifier we have is the URI, and I'm not sure that querying for that is the best way to go.

core/src/main/java/org/openmbee/sdvc/core/components/EventPublisher.java
HuiJun: Actually, this class isn't even used anymore. Removing

PR #46, opened 2020-02-14
Updates to allow TWC-based authentication delegation

BA-MBSE:
Updates to allow TWC-based authentication by reusing TWC tokens obtainable by MagicDraw plugins


.gitignore
```diff
@@ -26,3 +26,19 @@ localhost-env.json
 /dist/
 /nbdist/
 /.nb-gradle/
+
+
```
dlamoris, line 30: think **/bin/ should cover all of these

permissions/src/main/java/org/openmbee/sdvc/permissions/users/UserDetailsImpl.java
```diff
@@ -1,4 +1,4 @@
-package org.openmbee.sdvc.authenticator.security;
+package org.openmbee.sdvc.permissions.users;
```
dlamoris: think there may be a misunderstanding here - the UserDetailsImpl and UserDetailsServiceImpl deal only with authentication and not permissions - I think what can be misleading is the way we're leveraging the getGrantedAuthority interface - we're "hijacking" the GrantedAuthority to mean user groups (ex. ldap groups, or other kinds of groups, or it can be empty, depending on the auth and permission implementation combo). Since user permissions are per project/branch, there's no user permissions at the system level (besides admin, and the current implementations  add in 'mmsadmin' authority for mms system admins)

see other comments in the twc filter for more info

authenticator/authenticator.gradle
```diff
@@ -3,6 +3,7 @@ apply plugin: 'io.spring.convention.spring-module'
 dependencies {
     compile project(':rdb')
     compile project(':core')
+    compile project(':permissions')
```
dlamoris: should not be importing permissions (default permissions implementation), see other comments for more detail

twc/src/main/java/org/openmbee/sdvc/twc/security/TwcAuthenticationFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+		try {
+		    String associatedTWC = request.getHeader(TwcConstants.ASSOCIATED_TWC_HEADER);
+		    String token = request.getHeader(RestUtils.AUTHORIZATION);
+		    TwcAuthenticationProvider twcAuthProvider = twcConfig.getAuthNProvider(associatedTWC);
+		    if(twcAuthProvider != null) {
+                String authenticatedUser = twcAuthProvider.getAuthentication(token);
+                if(authenticatedUser != null) {
+                    UserDetailsImpl userDetails = this.userDetailsService.loadUserByUsername(authenticatedUser);
```
dlamoris: I don't think there's a need to use the userDetailsService here - that particular implementation assumes users are in the local database or usernames are synced to the local database on login (ex. see ldap module)
I think what you want is something like TwcUserDetailsImpl that implements Spring's UserDetails - if the decision to use UserDetailsImpl was because the /authentication controller was requiring it, that's something that's being fixed in a branch (see https://openmbee.atlassian.net/browse/MMS-53) 
(I also don't see twc authenticated users getting created/synced to the local db, so using this service shouldn't work)

The getGrantedAuthorities interface should return a list of SimpleGrantedAuthority (basically just strings) - this gets passed to the PermissionService interfaces down the line. The default implementation in the permissions module treats these as groups in the db the user belongs to - so permissions can be set for groups. Since PermissionService is an interface, there can be a TwcPermissionService that can interpret these passed in 'groups' however it wants, or it can ignore them. 

If a TwcPermissionService is implemented, then the permissions module should not be imported by the application, otherwise the default behavior is to lookup permissions set in the local db

dlamoris (changes requested):
also see https://openmbee.atlassian.net/browse/MMS-30 that's relevant to where security configs should live

permissions/src/main/java/org/openmbee/sdvc/permissions/users/UserDetailsImpl.java
BA-MBSE: We were trying to follow the same pattern without code duplication, but point taken.  Updated accordingly

twc/src/main/java/org/openmbee/sdvc/twc/security/TwcAuthenticationFilter.java
BA-MBSE: Our use cases require 

- authentication from multiple sources (From within Cameo, from a SSO provider, etc)
- projects from multiple sources (Cameo and other)

These all need to work together, so our goal is to decouple the Authentication and Authorization.  
There could be multiple sources of Authentication (TWC would just be one of them).   
The source of Authorization for each project/branch would be defined by project/branch attributes.  
Some projects could use the internal permissions scheme, and others could delegate to TWC for Authorization.

dlamoris: Right, for the first one (multiple authentication sources), that's already provided by spring security's filter and auth provider mechanism - it'll go through all filters/providers until one of them sets the authentication object - I'm just confused by the TwcAuthenticationFilter as to why after getting the user from the TWC Auth token, it's loading the user from the local database? This would only succeed if the local database gets synced with TWC users somehow, but I don't see that happening anywhere.

For the permissions - it'd be possible to do something similar I think, but it's not there yet - basically go through all possible permissionService implementations until it provides an answer - or have an implementation that does what you describe

BA-MBSE: I think the user will still need to exist locally.  If the user doesn't exist locally, but is authenticated by TWC, we should probably add them automatically.  I'll add that in.

I'm still thinking about how to integrate the permissions, but I like the idea of multiple permission service implementations.

dlamoris: Well, it's not strictly necessary for the user to exist locally for auth - that may be a convenience if we want to add user profiles/settings down the line - but that would be up to you - be aware though if you load the user details from the db it'll have no authorities unless you also sync those (if there are any that should be attached), otherwise you can also just return them from the TwcUserDetailsImpl directly.

I guess I'm trying to see if syncing the user in the local db is really what you want? It's not a requirement for auth

BA-MBSE: There is no need to sync the authorities with TWC since we're going to delegate those back to TWC anyway.  I agree that it can still work without adding the user to the db, but I think we still want to add them for the sake of consistency.  We will still want to pick up MMSADMIN rights for users that authenticate via TWC tokens

dlamoris (approved):
at some point we'll want to go through and clean up dependencies as part of https://openmbee.atlassian.net/browse/MMS-30

PR #50, opened 2020-02-26
Removing all equals overrides in subclasses

HuiJun:


data/src/main/java/org/openmbee/sdvc/data/domains/global/Base.java
```diff
@@ -58,8 +59,53 @@ public void setModified(Instant modified) {
...
+        try {
+            Field[] fields = getClass().getDeclaredFields();
+            for (Field field : fields) {
+                Object current = field.get(this);
+                Object value = field.get(o);
+                switch(field.getType().toString()) {
+                    case "String" :
+                    case "DecimalFormat" :
+                    case "BigDecimal":
+                        if (!current.equals(value)) {
```
dlamoris: what if current is null?

HuiJun: then npe

data/src/main/java/org/openmbee/sdvc/data/domains/global/Base.java
```diff
@@ -58,8 +59,53 @@ public void setModified(Instant modified) {
...
+                Object current = field.get(this);
+                Object value = field.get(o);
+                switch(field.getType().toString()) {
+                    case "String" :
+                    case "DecimalFormat" :
+                    case "BigDecimal":
+                        if ((current != null && !current.equals(value)) || (current == null && value != null)) {
+                            return false;
+                        }
+
```
dlamoris, line 85: think a break is needed here, otherwise current.equals(value) can be true but it'll fall through to default and potentially return false

data/src/main/java/org/openmbee/sdvc/data/domains/global/Base.java
```diff
@@ -58,8 +59,53 @@ public void setModified(Instant modified) {
...
     @Override
     public int hashCode() {
-        return (id == null) ? -1 : id.hashCode();
+        int seed = 42;
+        int hash = 0;
+        Field[] fields = getClass().getDeclaredFields();
+        for (Field field : fields) {
+            try {
+                hash = hash + (seed * field.get(this).hashCode());
```
dlamoris: also need null guard

data/src/main/java/org/openmbee/sdvc/data/domains/global/Base.java
HuiJun: Wouldn't we want it to return false if the loop goes to default and current != value?

PR #53, opened 2020-03-06
Permissions Delegation

BA-MBSE:
- adds ability to allow different permissions implementations for different projects/branches
- starting to add in TWC permissions delegation

Early pull request for review/feedback


rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/TWCIntegrationDAOImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import org.openmbee.sdvc.data.domains.global.TWCIntegration;
+import org.openmbee.sdvc.rdb.config.DatabaseDefinitionService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+import java.util.Optional;
+
+@Repository
+public class TWCIntegrationDAOImpl implements TWCIntegrationDAO {
```
dlamoris: this is just from a preliminary glance of the pr, but we're reluctant to add project specific tables, is the need just to store more metadata from the project json? we have an open jira for saving the entire project/ref jsons (https://openmbee.atlassian.net/browse/MMS-56) that should eliminate the need for this, (there's also an existing Metadata table connected to Project, but we haven't decided which route to take yet)

BA-MBSE: We missed the metadata table somehow.  Will refactor to use the metadata instead.

BA-MBSE:
@dlamoris Doris, are you ok with merging this as-is?  It's going to take a little longer than I expected to finish out the TWC interaction part.  If you're ok with filling that stuff in later, I'll take this off WIP

permissions/src/main/java/org/openmbee/sdvc/permissions/DefaultPermissionService.java
```diff
@@ -1,623 +1,213 @@
...
+import org.openmbee.sdvc.core.delegation.PermissionsDelegate;
 import org.openmbee.sdvc.permissions.exceptions.PermissionException;
 import org.openmbee.sdvc.rdb.repositories.*;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.data.util.Pair;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 
 @Service("defaultPermissionService")
 public class DefaultPermissionService implements PermissionService {
```
dlamoris, line 27: The permission service interface was originally intended to be the only abstraction, since now there's the delegation factory abstraction, i think it's worth looking at in the future if this should just become a core "PermissionService" impl instead of having the default qualifier (and move to core or something)

permissions/src/main/java/org/openmbee/sdvc/permissions/delegation/AbstractDefaultPermissionsDelegate.java
```diff
@@ -0,0 +1,31 @@
...
+import org.openmbee.sdvc.data.domains.global.Role;
+import org.openmbee.sdvc.rdb.repositories.GroupRepository;
+import org.openmbee.sdvc.rdb.repositories.RoleRepository;
+import org.springframework.data.util.Pair;
+
+import java.util.Optional;
+
+public abstract class AbstractDefaultPermissionsDelegate implements PermissionsDelegate {
+
+    abstract protected GroupRepository getGroupRepo();
```
dlamoris: It seems like a lot of the duplicate autowires in the subclasses can just be autowired in this class? then there's no need for the abstract getGroupRepo and getRoleRepo. Also private field injection is generally not recommended, so using setter injection would be better
https://blog.marcnuri.com/field-injection-is-not-recommended/

dlamoris (changes requested):

permissions/src/main/java/org/openmbee/sdvc/permissions/delegation/AbstractDefaultPermissionsDelegate.java
BA-MBSE: I updated it to have the common repositories in the base class and to use setter injection

permissions/src/main/java/org/openmbee/sdvc/permissions/DefaultPermissionService.java
BA-MBSE: Yeah, I was thinking the same thing.  Do you want to leave it as it is for now or go ahead and move it?

dlamoris: can leave it for now, there's probably more stuff in the wrong place, want to just go through and move things around once, i'll add a jira to remember

dlamoris (approved):

PR #54, opened 2020-03-09
MMS-60 Added holding bin and view instances bin

BA-MBSE:


example/sdvc.postman_collection.json
```diff
@@ -193,6 +193,55 @@
 					},
 					"response": []
 				},
+				{
```
dlamoris: it may be better to leave this collection as is and add cameo specific tests to the cameo collection json (there's one already for testing mounts) - i've wanted to separate out the tests into more logical collections but haven't made an issue yet. 
To fix this collection so it's not making a cameo project - change the project posts json to have projectType: default (i think leaving it out would also work)

cameo/src/main/java/org/openmbee/sdvc/cameo/services/CameoProjectService.java
```diff
@@ -1,11 +1,78 @@
...
+        e.setCreated(projectJson.getCreated());
+        e.setModifier(projectJson.getModifier());
+        e.setModified(projectJson.getModified());
+        e.setName(name);
+        e.put(CameoConstants.OWNERID, projectJson.getProjectId());
+        e.put(CameoConstants.TYPE, CameoConstants.PACKAGE_TYPE);
+        e.put(CameoConstants.ISGROUP, false);
+        e.put(CameoConstants.DOCUMENTATION, "");
+        e.put(CameoConstants.VISIBILITY, CameoConstants.PUBLIC_VISIBILITY);
+        //TODO Which fields need to be filled in?  Commented out blank ones from donbot src
```
dlamoris: i'll need to check back how mdk will process them, generally it should just import it and md will make it valid on creation but that'll incur a diff on mdk

dlamoris (changes requested):

cameo/src/main/java/org/openmbee/sdvc/cameo/services/CameoProjectService.java
BA-MBSE: Ok, I removed the empty fields.  My team is planning to work through updating MDK anyway and we'll pick up and fix incompatibilities then.

example/sdvc.postman_collection.json
BA-MBSE: Changed the sdvc collection to use 'default' projects instead of 'cameo' to avoid the errors.  Added a test for the default elements to the cameo collection

dlamoris (approved):

PR #55, opened 2020-03-12
Mms-50 Allow password changes

BA-MBSE:


authenticator/src/main/java/org/openmbee/sdvc/authenticator/security/UserDetailsServiceImpl.java
```diff
@@ -49,4 +49,15 @@ public User register(String username, String password, boolean isAdmin) {
...
     }
 
+    @Transactional
+    public void changeUserPassword(String username, String password) {
+        Optional<User> userOptional = userRepository.findByUsername(username);
+        if(! userOptional.isPresent()) {
+                throw new UsernameNotFoundException(
+                    String.format("No user found with username '%s'.", username));
+        }
+        User user = userOptional.get();
```
dlamoris, line 61: think we'll want to check if the existing user password is null/empty and throw a rejection since that means the user is a remote user (ldap/twc) and password change shouldn't be allowed

dlamoris (changes requested):

authenticator/src/main/java/org/openmbee/sdvc/authenticator/security/UserDetailsServiceImpl.java
BA-MBSE: Do you think we should still allow admins to set a password if the user is an ldap/twc user?

dlamoris: I think as of now, no, but
that does open the question if we want to allow converting an "external" user to local user down the line...we currently don't have that use case, if that's a real need we might want to add a dedicated column to user that specifies local vs external, that might be a separate pr with all the other cases that goes with it

PR #56, opened 2020-03-12
MMS-33 Reject invalid project ids

BA-MBSE:
Added regular expression check to project ids to enforce only alpha-numeric with underscores and dashes
Updated database create to work with dashes
gradle update was to fix jacoco/java11 error, not sure why this module had this problem.

### DIFF
--- a/crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
+++ b/crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
@@ -35,7 +35,7 @@ import org.springframework.web.bind.annotation.RestController;
 @RequestMapping("/projects")
 public class ProjectsController extends BaseController {
 
-    private static final String PROJECT_ID_VALID_PATTERN = "^[\\w-]{1,62}$";
+    private static final String PROJECT_ID_VALID_PATTERN = "^[\\w-]+$";
 
     ProjectDAO projectRepository;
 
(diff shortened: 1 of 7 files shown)


crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
```diff
@@ -35,6 +35,8 @@
 @RequestMapping("/projects")
 public class ProjectsController extends BaseController {
 
+    private static final String PROJECT_ID_VALID_PATTERN = "^[\\w-]{1,62}$";
```
dlamoris: the 62 limit is a side effect of how we implemented the db layer, in reality some twc server projects's id can routinely exceed that limit (at least what we have), so I'm not sure about limiting it here. (the db will automatically truncate too long ids) So far what's left from the truncation are still enough to be unique in the db. I think having a note/warning should be enough instead of rejecting in the code, or if the implementation layer has further restrictions, to throw exceptions at the impl layer instead.

dlamoris (changes requested):
can you also add the check for branch creation?

crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
BA-MBSE: Updated to not limit length + added check from branch ids

PR #59, opened 2020-04-02
Mms 36 permissions response

BA-MBSE:
Builds on #53


data/src/main/java/org/openmbee/sdvc/data/domains/global/Branch.java
```diff
@@ -75,4 +76,19 @@ public boolean isInherit() {
     public void setInherit(boolean inherit) {
         this.inherit = inherit;
     }
+
+    @Override
+    public boolean equals(Object o) {
```
dlamoris: there's a generic equals and hashCode in Base, would prefer to try fixing that if it's not adequate instead of adding them to all subclasses
if it's not possible to do in Base, should just delete the ones in Base (might be issues with hibernate proxies? or some fields should be skipped if they're collections...)

BA-MBSE: Yeah, I'm not sure what side effects these will have since they don't all include all of the fields. These were only meant to get the desired behavior from the recalculateInhertiedPerms() methods.  I can reimplement it a different way.

permissions/src/main/java/org/openmbee/sdvc/permissions/DefaultPermissionService.java
```diff
@@ -82,102 +86,122 @@ public void initBranchPerms(String projectId, String branchId, boolean inherit,
...
-        permissionsDelegate.updateGroupPermissions(req);
+        return permissionsDelegate.updateGroupPermissions(req);
     }
 
     @Override
     @Transactional
-    public void setProjectInherit(boolean isInherit, String projectId) {
+    public PermissionUpdatesResponse setProjectInherit(boolean isInherit, String projectId) {
+        PermissionUpdatesResponseBuilder responseBuilder = new PermissionUpdatesResponseBuilder();
+        responseBuilder.setInherit(true);
```
dlamoris, line 167: there's several places where setInherit is hardcoded to true, think it should just be the parameter isInherit? Same for setOrg or setProject public (the method name is a bit misleading since it's actually setting it based on an argument

dlamoris: actually if it's just to keep track of whether it's been set/processed then nevermind

core/src/main/java/org/openmbee/sdvc/core/builders/PermissionUpdateResponseBuilder.java
```diff
@@ -0,0 +1,155 @@
+package org.openmbee.sdvc.core.builders;
+
+import org.openmbee.sdvc.core.objects.PermissionUpdateResponse;
+import org.openmbee.sdvc.data.domains.global.*;
+
+import java.util.*;
+
+public class PermissionUpdateResponseBuilder {
+
+    private class PermissionUpdateWrapper {
```
dlamoris, line 10: maybe the hash and equals can be done in PermissionUpdate in this case so the wrapper won't be needed

PR #62, opened 2020-04-14
Feature/openapi hardening

dlamoris:


example/openapi/api-docs.yaml
```diff
@@ -1035,331 +1361,458 @@ components:
...
+          readOnly: true
+        _commitId:
           type: string
-        _created:
+          readOnly: true
+      additionalProperties: true
+    CommitUpdatedJson:
+      type: object
+      properties:
+        previousDocId:
```
ivan-gomes: `previousDocId` doesn't start with an underscore but `_docId` does. Also applies to `CommitDeletedJson#previousDocId`.

example/openapi/api-docs.yaml
```diff
@@ -1035,331 +1361,458 @@ components:
...
           type: string
-        commits:
+          nullable: true
+        refs:
           type: array
           items:
-            $ref: '#/components/schemas/CommitJson'
-    ProjectJson:
+            $ref: '#/components/schemas/RefJson'
+    CommitAddedJson:
```
ivan-gomes: Can we merge the `Commit + (Added|Updated|Deleted) + Json` models? None of the properties are specified as required, so CommitUpdatedJson (renamed) can be used for all of them.

ivan-gomes (changes requested):

example/openapi/api-docs.yaml
dlamoris: _ added

example/openapi/api-docs.yaml
dlamoris: changed to ElementUpdated

example/openapi/api-docs.yaml
```diff
@@ -1035,331 +1412,419 @@ components:
...
+    CommitsRequest:
+      required:
+      - commits
+      type: object
+      properties:
+        commits:
+          type: array
+          items:
+            $ref: '#/components/schemas/Commit'
+    ElementUpdated:
```
ivan-gomes: I think `ElementVersion` (consistent with SysML v2), `ElementChange`, `ElementRevision`, etc. may be a better name as update is usually connoted as disjoint from create/delete (CRUD).

ivan-gomes (changes requested):

example/openapi/api-docs.yaml
dlamoris: changed to ElementVersion

PR #63, opened 2020-04-15
Changes to allow dashes "-" in branch names

BA-MBSE:
+ updated to latest JPA/hibernate
The major change was to set hibernate.globally_quoted_identifiers=true, but this has some side effects
1. column names are now case sensitive
2. smallint type can't be used (The PG dialect doesn't recognize it as a type - or something like that - so hibernate wraps it in quotes and breaks it)

### DIFF
--- a/core/src/main/java/org/openmbee/sdvc/core/config/ContextObject.java
+++ b/core/src/main/java/org/openmbee/sdvc/core/config/ContextObject.java
@@ -70,6 +70,6 @@ public class ContextObject {
     }
 
     private void branchIdUpdated() {
-        this.dbTableSuffix = this.branchId.equals(MASTER_BRANCH) ? "" : this.branchId.replace('-', '_');
+        this.dbTableSuffix = this.branchId.equals(MASTER_BRANCH) ? "" : this.branchId;
     }
 }
--- a/example/src/main/resources/application.properties.example
+++ b/example/src/main/resources/application.properties.example
@@ -30,6 +30,7 @@ spring.datasource.initialization-mode=always
 
 # The SQL dialect makes Hibernate generate better SQL for the chosen database
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQL94Dialect
+spring.jpa.properties.hibernate.globally_quoted_identifiers=true
 #spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL57Dialect
 #spring.jpa.properties.hibernate.dialect.storage_engine=innodb
 
--- a/rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
+++ b/rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
@@ -34,11 +34,11 @@ import org.springframework.stereotype.Service;
 public class DatabaseDefinitionService {
 
     private static final Pattern pattern = Pattern.compile("\\$\\{(.+?)\\}");
-    private static final String COPY_SQL = "INSERT INTO %s SELECT * FROM %s";
-    private static final String COPY_IDX = "SELECT SETVAL('%s_id_seq', COALESCE((SELECT MAX(id) FROM %s), 1), true)";
+    private static final String COPY_SQL = "INSERT INTO \"%s\" SELECT * FROM \"%s\"";
+    private static final String COPY_IDX = "SELECT SETVAL('%s_id_seq', COALESCE((SELECT MAX(\"id\") FROM \"%s\"), 1), true)";
 
-    private static final String INITIAL_PROJECT = "INSERT INTO nodes (id, nodeid, docid, initialcommit, lastcommit, nodetype, deleted) VALUES (0, ?, ?, ?, ?, ?, false)";
-    private static final String INITIAL_REF = "INSERT INTO branches (id, branchid, branchname, tag, deleted, timestamp) VALUES (0, 'master', 'master', false, false, NOW());";
+    private static final String INITIAL_PROJECT = "INSERT INTO \"nodes\" (\"id\", \"nodeId\", \"docId\", \"initialCommit\", \"lastCommit\", \"nodeType\", \"deleted\") VALUES (0, ?, ?, ?, ?, ?, false)";
+    private static final String INITIAL_REF = "INSERT INTO \"branches\" (\"id\", \"branchId\", \"branchName\", \"tag\", \"deleted\", \"timestamp\") VALUES (0, 'master', 'master', false, false, NOW());";
 
     protected final Logger logger = LogManager.getLogger(getClass());
     private CrudDataSources crudDataSources;
@@ -192,6 +192,7 @@ public class DatabaseDefinitionService {
         properties.put("hibernate.physical_naming_strategy",
             "org.openmbee.sdvc.rdb.config.SuffixedPhysicalNamingStrategy");
 
+        properties.put("hibernate.globally_quoted_identifiers", true);
         return properties;
     }
 }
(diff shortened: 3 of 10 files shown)


BA-MBSE:
Still need to add postman tests

core/src/main/java/org/openmbee/sdvc/core/config/ContextObject.java
```diff
@@ -70,6 +70,6 @@ private void projectIdUpdated() {
     }
 
     private void branchIdUpdated() {
-        this.dbTableSuffix = this.branchId.equals(MASTER_BRANCH) ? "" : this.branchId.replace('-', '_');
+        this.dbTableSuffix = this.branchId.equals(MASTER_BRANCH) ? "" : this.branchId;
```
dlamoris, line 73: this probably don't need to change with the quotes added now, would need to be tested..

example/src/main/resources/application.properties.example
```diff
@@ -30,6 +30,7 @@ spring.datasource.initialization-mode=always
 
 # The SQL dialect makes Hibernate generate better SQL for the chosen database
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQL94Dialect
+spring.jpa.properties.hibernate.globally_quoted_identifiers=true
```
dlamoris, line 33: i would say this isn't needed, the things that this would affect are things we already define, the side effects that comes with it aren't really worth it

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
```diff
@@ -34,11 +34,11 @@
...
 
     private static final Pattern pattern = Pattern.compile("\\$\\{(.+?)\\}");
-    private static final String COPY_SQL = "INSERT INTO %s SELECT * FROM %s";
-    private static final String COPY_IDX = "SELECT SETVAL('%s_id_seq', COALESCE((SELECT MAX(id) FROM %s), 1), true)";
+    private static final String COPY_SQL = "INSERT INTO \"%s\" SELECT * FROM \"%s\"";
+    private static final String COPY_IDX = "SELECT SETVAL('%s_id_seq', COALESCE((SELECT MAX(\"id\") FROM \"%s\"), 1), true)";
 
-    private static final String INITIAL_PROJECT = "INSERT INTO nodes (id, nodeid, docid, initialcommit, lastcommit, nodetype, deleted) VALUES (0, ?, ?, ?, ?, ?, false)";
-    private static final String INITIAL_REF = "INSERT INTO branches (id, branchid, branchname, tag, deleted, timestamp) VALUES (0, 'master', 'master', false, false, NOW());";
+    private static final String INITIAL_PROJECT = "INSERT INTO \"nodes\" (\"id\", \"nodeId\", \"docId\", \"initialCommit\", \"lastCommit\", \"nodeType\", \"deleted\") VALUES (0, ?, ?, ?, ?, ?, false)";
```
dlamoris, line 40: don't think we need quotes here since this is for the master node table, same for the master branch below

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
```diff
@@ -192,6 +192,7 @@ public void copyTablesFromParent(String target, String parent, String parentComm
         properties.put("hibernate.physical_naming_strategy",
             "org.openmbee.sdvc.rdb.config.SuffixedPhysicalNamingStrategy");
 
+        properties.put("hibernate.globally_quoted_identifiers", true);
```
dlamoris, line 195: don't think we need it

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/branch/BranchDAOImpl.java
```diff
@@ -27,8 +27,8 @@ public void setBranchesOperations(DatabaseDefinitionService branchesOperations)
         this.branchesOperations = branchesOperations;
     }
 
-    private final String INSERT_SQL = "INSERT INTO branches (description, branchId, branchName, parentRefId, parentCommit, timestamp, tag, deleted) VALUES (?, ?, ?, ?, ?, ?, ?, ?)";
-    private final String UPDATE_SQL = "UPDATE branches SET description = ?, branchId = ?, branchName = ?, parentRefId = ?, parentCommit = ?, timestamp = ?, tag = ?, deleted = ? WHERE id = ?";
+    private final String INSERT_SQL = "INSERT INTO \"branches\" (\"description\", \"branchId\", \"branchName\", \"parentRefId\", \"parentCommit\", \"timestamp\", \"tag\", \"deleted\") VALUES (?, ?, ?, ?, ?, ?, ?, ?)";
```
dlamoris, line 30: I don't think we need to quote anything in the DAOImpls except for %s in NodeDAOImpl - that's the only thing that would change with different branches

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/node/NodeDAOImpl.java
```diff
@@ -21,8 +21,8 @@
 @Component
 public class NodeDAOImpl extends BaseDAOImpl implements NodeDAO {
 
-    private final String INSERT_SQL = "INSERT INTO nodes%s (nodeid, docid, lastcommit, initialcommit, deleted, nodetype) VALUES (?, ?, ?, ?, ?, ?)";
-    private final String UPDATE_SQL = "UPDATE nodes%s SET nodeid = ?, docid = ?, lastcommit = ?, initialcommit = ?, deleted = ?, nodetype = ? WHERE id = ?";
+    private final String INSERT_SQL = "INSERT INTO \"nodes%s\" (\"nodeId\", \"docId\", \"lastCommit\", \"initialCommit\", \"deleted\", \"nodeType\") VALUES (?, ?, ?, ?, ?, ?)";
```
dlamoris, line 24: only the nodes%s should need quotes

dlamoris (changes requested):

example/src/main/resources/application.properties.example
BA-MBSE: The limitation this works around is in DatabaseDefinitionService.generateBranchSchemaFromModels().  Here, hibernate DDL is creating the Nodes* tables before the select into gets to run.  Hibernate can't create a table with a dash without this set in DatabaseDefinitionService.getSchemaProperties().

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
BA-MBSE: We might not need it.  I was just trying to make things look consistent.

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/branch/BranchDAOImpl.java
BA-MBSE: I couldn't get things to work without everything quoted.  It's like if one identifier is quoted, it expects them all to be.

example/src/main/resources/application.properties.example
dlamoris: ah, I remember this now - I also commented on the jira - there's a SuffixPhysicalNamingStrategy that we can set this for just table names instead of doing it for everything

rdb/src/main/java/org/openmbee/sdvc/rdb/repositories/branch/BranchDAOImpl.java
dlamoris: when the global quote property quotes everything, it made the column names case sensitive which then required everything to be quoted (otherwise i believe even if you put something like branchId the db would normalize it to branchid) if we just quote the table on creation in the Suffix strategy then only the nodes table should need to quoted

example/src/main/resources/application.properties.example
dlamoris: actually would be better if it's just set for nodes* tables and just leave other tables as is

BA-MBSE:
Superseded by #64 - Closing

PR #64, opened 2020-04-20
Changes to allow dashes "-" in branch names

BA-MBSE:
Added quoting to the nodes tables so hibernate properly handles table names with dashes

Still need to add test cases


BA-MBSE:
Added tests as part of refactoring for #65

PR #65, opened 2020-04-21
MMS-79 - Logically separating test collections + Tests for MMS-87 - dashes in branch names

BA-MBSE:
Split sdvc.postman_collection.json into crud.postman_collection.json and permissions.postman_collection.json

* Added tests (MMS-87) for dashes in org, project, and branch names as part of the refactoring of permissions.postman_collection.json
* Fixed bug in DefaultPermissionService.java


dlamoris (changes requested):
can you copy the "login using admin" step to be the first step in permissions tests so it can be run independently?
also a lot of the tests inherit auth from the collection itself, there should be something like this:
```
"auth": {
		"type": "bearer",
		"bearer": [
			{
				"key": "token",
				"value": "{{token}}",
				"type": "string"
			}
		]
	},
```
from the original sdvc collection that needs to be copied over to both new collections (the "login using admin" step will set the token to be used in the rest of the tests)

dlamoris:
this will have conflicts with pr #59 since there's a fix in the tests there...seems it's easier if i merge this first and redo the fixes from 59 in the other pr?

BA-MBSE:
Whichever way you want to work the conflicts is fine with me.  I don't mind fixing the errors if you want to merge 59 first.  
And, ok I'll fix the tests so we're using tokens.

dlamoris:
I'll merge this first, since the other test changes has less to refix

PR #69, opened 2020-04-29
MMS-99 - Basic Search Capabilities to support MDK work

BA-MBSE:
Adding in some search capability to allow Cameo MDK work to move forward.  This will likely be replaced by the final graph interface


search/src/main/java/org/openmbee/sdvc/search/controllers/SearchController.java
```diff
@@ -0,0 +1,86 @@
...
+        @PathVariable String refId,
+        @RequestBody BasicSearchRequest request
+    ) {
+        ElementsResponse res = searchService.recursiveSearch(projectId, refId, request.getParams(), request.getRecurse());
+        handleSingleResponse(res);
+        return res;
+    }
+
+
+    private void handleSingleResponse(BaseResponse res) {
```
dlamoris: this was just for single element gets returning the right http response code, i don't think it's needed for search since there won't be any rejections

elastic/src/main/java/org/openmbee/sdvc/elastic/services/ElasticSearchService.java
```diff
@@ -0,0 +1,151 @@
...
+        return recursiveParams;
+    }
+
+
+    private List<ElementJson> doSearch(List<Node> validNodes, Map<String, String> params) throws IOException {
+        List<ElementJson> result = new ArrayList<>();
+
+        SearchRequest searchRequest = new SearchRequest(Index.NODE.get());
+        BoolQueryBuilder query = QueryBuilders.boolQuery();
+        query.filter(QueryBuilders.termsQuery(ElementJson.DOCID, validNodes.stream().map(Node::getDocId).toArray()));
```
dlamoris: There's performance degradation the more terms are put in for a filter - for testing should be fine, but once the amount of elements in a project reach a certain size (like 100000) it's better to split this up (the default elastic limit on terms is 65536)
This can also be replaced by a `QueryBuilders.termQuery(ElementJson.INREFIDS, refid)`, and then just filter on the docids in memory just in case

search/src/main/java/org/openmbee/sdvc/search/controllers/SearchController.java
```diff
@@ -0,0 +1,86 @@
...
+    private MethodSecurityService mss;
+    private SearchService searchService;
+
+    @Autowired
+    public void setSearchService(SearchService searchService) {
+        this.searchService = searchService;
+    }
+
+    @Autowired
+    public void setMss(MethodSecurityService mss) {
```
dlamoris: no need to inject mss, the preauthorize will already look it up as a service

example/search.postman_collection.json
```diff
@@ -0,0 +1,416 @@
...
+		},
+		{
+			"name": "add org s",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "10fb6974-d6a5-4c9f-a4d8-1fc0c6739c33",
+						"exec": [
+							"pm.test(\"response has org a\", function () {",
```
dlamoris: fix test to check for 's'

example/search.postman_collection.json
```diff
@@ -0,0 +1,416 @@
...
+				"header": [
+					{
+						"key": "Content-Type",
+						"value": "application/json",
+						"type": "text"
+					}
+				],
+				"body": {
+					"mode": "raw",
+					"raw": "{\n\t\"projects\": [\n\t\t{\n\t\t\t\"id\": \"ss\", \n\t\t\t\"name\": \"ss\",\n\t\t\t\"orgId\": \"s\",\n\t\t\t\"schema\": \"default\"\n\t\t}\n\t]\n}"
```
dlamoris: it might be better to use cameo schema here to ensure the test search is using keywords..

example/search.postman_collection.json
```diff
@@ -0,0 +1,416 @@
...
+			"name": "add elements to ss",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "f23976f1-afb8-495f-892a-dcbcc7360665",
+						"exec": [
+							"pm.test(\"response has elements\", function () {",
+							"    var jsonData = pm.response.json();",
+							"    pm.expect(jsonData.elements.length).to.eql(3);",
```
dlamoris: fix test to check for 4 elements instead of 3

example/search.postman_collection.json
```diff
@@ -0,0 +1,416 @@
...
+		},
+		{
+			"name": "search for x again",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"id": "4ea3441a-8a0f-421a-ba27-67ac1351775d",
+						"exec": [
+							"pm.test(\"response has 1 element\", function () {",
```
dlamoris: fix test description to say 0 element

example/search.postman_collection.json
```diff
@@ -0,0 +1,416 @@
...
+			],
+			"request": {
+				"method": "GET",
+				"header": [],
+				"body": {
+					"mode": "raw",
+					"raw": ""
+				},
+				"url": {
+					"raw": "{{host}}/projects/ss/refs/master/search?name=x",
```
dlamoris, line 221: this is dependent on the search impl, but as of elastic, searching for name would most likely not give 'exact' results since names aren't keyword and are tokenized (the single letter just happens to work)

elastic/src/main/java/org/openmbee/sdvc/elastic/services/ElasticSearchService.java
```diff
@@ -0,0 +1,151 @@
...
+
+        for(Map.Entry<String, String> e : params.entrySet()) {
+            query.filter(QueryBuilders.termQuery(e.getKey(), e.getValue()));
```
dlamoris, line 116: fyi when testing, this would only work for things with keyword mapping, anything text or paragraphs would need a `match` query which is outside the scope of this pr

dlamoris (changes requested):

search/src/main/java/org/openmbee/sdvc/search/controllers/SearchController.java
BA-MBSE: Removed, thanks

elastic/src/main/java/org/openmbee/sdvc/elastic/services/ElasticSearchService.java
BA-MBSE: Refactored as suggested.

search/src/main/java/org/openmbee/sdvc/search/controllers/SearchController.java
BA-MBSE: Hmm, didn't know that.  Thanks!

example/search.postman_collection.json
BA-MBSE: Updated

example/search.postman_collection.json
BA-MBSE: Fixed

example/search.postman_collection.json
BA-MBSE: Good point. updated to use id

PR #74, opened 2020-05-04
MMS-98 - Artifact Storage

BA-MBSE:


artifacts/src/main/java/org/openmbee/sdvc/artifacts/controllers/ArtifactController.java
```diff
@@ -0,0 +1,86 @@
...
+        @PathVariable String refId,
+        @PathVariable String elementId,
+        @PathVariable String extension,
+        @RequestParam(required = false) Map<String, String> params,
+        Authentication auth) {
+
+        params.put(ArtifactConstants.EXTENSION_PARAM, extension);
+        ArtifactResponse artifact = artifactService.get(projectId, refId, elementId, params);
+
+        if(artifact == null) {
```
dlamoris: instead of checking for null here, would be better if the artifact service call just throws an exception so it'll be caught by the controller advice

crud/src/main/java/org/openmbee/sdvc/crud/controllers/elements/ElementsController.java
```diff
@@ -71,6 +71,8 @@ public ElementsResponse createOrUpdateElements(
         throw new BadRequestException(response.addMessage("Empty"));
     }
 
+
```
dlamoris: comment is for the /elementId getElement - if you add produces = MediaType.APPLICATION_JSON_VALUE, it won't throw an error if there's a same mapping with no produces in artifacts controller

example/src/main/java/org/openmbee/sdvc/example/config/ExampleSecurityConfig.java
```diff
@@ -96,4 +103,11 @@ private CorsFilter corsFilter() {
         source.registerCorsConfiguration("/**", config);
         return new CorsFilter(source);
     }
+
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        if(interceptors != null) {
+            interceptors.forEach(v -> v.register(registry));
+        }
+    }
```
dlamoris, line 109: the interceptor shouldn't be needed if content negotiation is used - https://www.baeldung.com/spring-mvc-content-negotiation-json-xml
something like this
```
@Override
    public void configureContentNegotiation(ContentNegotiationConfigurer configurer) {
        configurer.favorPathExtension(false).
            favorParameter(false).
            ignoreAcceptHeader(false).
            useJaf(false).
            defaultContentType(MediaType.APPLICATION_JSON);
    }
```
this sends requests that don't have accept defined to default to json, along with adding `produces` to the elements get controller method, there can be another controller method for artifacts with the same url mapping (but no produces) so it can handle anything else with accept

artifacts/src/main/java/org/openmbee/sdvc/artifacts/service/DefaultArtifactService.java
```diff
@@ -0,0 +1,211 @@
...
+    @Autowired
+    public void setProjectRepository(ProjectDAO projectRepository) {
+        this.projectRepository = projectRepository;
+    }
+
+    @Override
+    public ArtifactResponse get(String projectId, String refId, String id, Map<String, String> params) {
+        NodeService nodeService = getNodeService(projectId);
+        ElementJson elementJson = getElement(nodeService, projectId, refId, id, params);
+        if(elementJson == null){
```
dlamoris: prefer throwing not found exceptions than returning null

artifacts/src/main/java/org/openmbee/sdvc/artifacts/service/DefaultArtifactService.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        String mimeType = getMimeTypeOfFile(file);
+        String fileExtension = getFileExtension(file);
+        String artifactLocation = artifactStorage.store(fileContents, elementJson, mimeType);
+
+        elementJson = attachOrUpdateArtifact(elementJson, artifactLocation, fileExtension, mimeType, "internal");
+        ElementsRequest elementsRequest = new ElementsRequest();
+        elementsRequest.setElements(Arrays.asList(elementJson));
+        nodeService.createOrUpdate(projectId, refId, elementsRequest, params, user);
```
dlamoris: can just return the response from nodeService, there's no guarantee the elementJson is updated in place with all the new metadata

artifacts/src/main/java/org/openmbee/sdvc/artifacts/service/DefaultArtifactService.java
```diff
@@ -0,0 +1,211 @@
...
+            throw new NotFoundException("Artifact not found");
+        }
+
+        ArtifactJson artifact = getExistingArtifact(artifacts, params);
+        if(artifact != null) {
+            artifacts.remove(artifact);
+            ArtifactJson.setArtifacts(elementJson, artifacts);
+            ElementsRequest elementsRequest = new ElementsRequest();
+            elementsRequest.setElements(Arrays.asList(elementJson));
+            nodeService.createOrUpdate(projectId, refId, elementsRequest, params, user);
```
dlamoris: just return from nodeService

artifacts/src/main/java/org/openmbee/sdvc/artifacts/json/ArtifactJson.java
```diff
@@ -0,0 +1,95 @@
...
+    public ArtifactJson setLocationType(String locationType) {
+        this.put(LOCATIONTYPE, locationType);
+        return  this;
+    }
+
+    public static List<ArtifactJson> getArtifacts(ElementJson elementJson){
+
+        List<Object> rawArtifacts = (List)elementJson.get(ARTIFACTS);
+        if(rawArtifacts == null || rawArtifacts.isEmpty()) {
+            return null;
```
dlamoris: prefer return empty list instead of null, so caller doesn't have to check for nulls

json/src/main/java/org/openmbee/sdvc/json/ElementJson.java
```diff
@@ -4,6 +4,8 @@
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.media.Schema.AccessMode;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
```
dlamoris: remove unused

dlamoris (changes requested):

crud/src/main/java/org/openmbee/sdvc/crud/controllers/elements/ElementsController.java
BA-MBSE: The story called for the elements controller to handle JSON and empty Accept headers. There's nothing to handle empty Accept headers if produces is defined and the artifacts module is not included in a build.

example/src/main/java/org/openmbee/sdvc/example/config/ExampleSecurityConfig.java
BA-MBSE: Ok, I'll look into this

crud/src/main/java/org/openmbee/sdvc/crud/controllers/elements/ElementsController.java
BA-MBSE: handled via content negotiation

artifacts/src/main/java/org/openmbee/sdvc/artifacts/service/DefaultArtifactService.java
BA-MBSE: now throwing instead

example/src/main/java/org/openmbee/sdvc/example/config/ExampleSecurityConfig.java
BA-MBSE: Reimplemented using this method.  Much cleaner now, thanks!

crud/src/main/java/org/openmbee/sdvc/crud/services/DefaultNodeService.java
```diff
@@ -130,6 +137,8 @@ public ElementsResponse createOrUpdate(String projectId, String refId, ElementsR
         ContextHolder.setContext(projectId, refId);
         boolean overwriteJson = Boolean.parseBoolean(params.get("overwrite"));
 
+        preCommitChanges(projectId, refId, req.getElements(), params, user);
```
dlamoris: Since the artifact subscriber is relying on checking for the artifact controller, this feels like it should really be called at the controller level - it'll basically be a hook for processing input going into /elements - the criteria for applying the processing is url based. The service interface also isn't guaranteed to be invoked through a request handler.

It may be worthwhile to also have a hook for dealing with input before it gets processed at the service level, but I think that would be something along the lines of a subclass override.

BA-MBSE: Reworked this into a generic EmbeddedHookService, and installed the hook at the controller level.

core/src/main/java/org/openmbee/sdvc/core/utils/RequestUtils.java
```diff
@@ -0,0 +1,52 @@
+package org.openmbee.sdvc.core.utils;
```
dlamoris: can remove since no longer needed

PR #77, opened 2020-05-26
Synchronizing differences between internal repository and open source

BA-MBSE:
+ Adding authentication delegation for teamwork cloud
+ Adding changes for MDK compatibility


core/core.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     api commonDependencies.'log4j'
     api commonDependencies.'log4j-api'
     api commonDependencies.'swagger-annotations'
+    api 'org.json:json:20190722'
```
HuiJun: This dependency and JsonUtils seems unnecessary, in core at least.

HuiJun (changes requested):

core/src/main/java/org/openmbee/sdvc/core/utils/RestUtils.java
```diff
@@ -33,4 +37,15 @@ public String getCookieValue(ResponseEntity<?> responseEntity, String cookieName
 
         return null;
     }
+    
+    
+    public HttpHeaders basicAuthHeader(String username, String password){
```
HuiJun, line 42: This may go against our stateless architecture design goals. Why do we need to set cookies? If necessary, can we move this where it's utilized, instead of in core?

HuiJun (changes requested):

core/src/main/java/org/openmbee/sdvc/core/utils/RestUtils.java
BA-MBSE: This is just setting a basic auth header.  No cookies are being set.
The cookie stuff in the previous method is pulling cookies out of a Response  - this is used for getting authentication details from Teamwork Cloud.  We can move it, but it seems generic enough for core.

core/core.gradle
BA-MBSE: Yeah, we can move it

core/src/main/java/org/openmbee/sdvc/core/utils/RestUtils.java
HuiJun: Sorry, I was referring to the file as a whole. I'm thinking we could move RestUtils to authenticator instead of having them in core.

dlamoris: I don't think twc imports the auth module - seems like they're only used by twc anyways, may be better to just move to twc

BA-MBSE: Ok, moved to twc module

HuiJun (approved):

PR #78, opened 2020-06-05
Adding ability to create mapping between commits and TWC revisions

BA-MBSE:
+ cleanup
+ Setting RDB JPA configuration as @Primary to allow for secondary data sources

### DIFF
--- a/twc/src/main/java/org/openmbee/sdvc/twc/twcrevisionmmscommitmap/TwcRevisionMmsCommitMapController.java
+++ b/twc/src/main/java/org/openmbee/sdvc/twc/twcrevisionmmscommitmap/TwcRevisionMmsCommitMapController.java
@@ -1,0 +1,29 @@
+package org.openmbee.sdvc.twc.twcrevisionmmscommitmap;
+
+import org.openmbee.sdvc.core.objects.CommitsResponse;
+import org.openmbee.sdvc.crud.controllers.BaseController;
+import org.openmbee.sdvc.twc.services.TwcRevisionMmsCommitMapService;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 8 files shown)


twc/src/main/java/org/openmbee/sdvc/twc/twcrevisionmmscommitmap/TwcRevisionMmsCommitMapController.java
```diff
@@ -0,0 +1,29 @@
...
+    private TwcRevisionMmsCommitMapService twcRevisionMmsCommitMapService;
+
+    @Autowired
+    public void setTwcRevisionMmsCommitMapService(TwcRevisionMmsCommitMapService twcRevisionMmsCommitMapService) {
+        this.twcRevisionMmsCommitMapService = twcRevisionMmsCommitMapService;
+    }
+
+    @PutMapping(value = "/twc-revision/{revisionId}")
+    @PreAuthorize("@mss.hasProjectPrivilege(authentication, #projectId, 'PROJECT_EDIT', true)")
+    public CommitsResponse updateTwcRevisionID(
```
dlamoris, line 23: does the key have to be twc-revisionId in the commit object or associated separately from element posts? there's currently 2 keys 'comment' and 'source' that are transferred over from element posts that can be used also, just fyi

dlamoris: oh, i guess you don't know the twc revision id until after the elements are posted to mms...

dlamoris (approved):

HuiJun (approved):

PR #79, opened 2020-06-12
Added ability to get TWC revisions associated with commits of a branch

BA-MBSE:


twc/src/main/java/org/openmbee/sdvc/twc/services/TwcRevisionMmsCommitMapService.java
```diff
@@ -35,4 +57,38 @@ public CommitsResponse updateTwcRevisionID(String projectId, String commitId, St
...
+    public List<CommitJson> getTwcRevisionList(String projectId, String refId) {
+        List<CommitJson> commits = new ArrayList<>();
+        ContextHolder.setContext(projectId);
+        Optional<Branch> ref = branchRepository.findByBranchId(refId);
+        if (!ref.isPresent()) {
+            throw new NotFoundException("Branch not found");
+        }
+        try {
+            List<Commit> refCommits = commitRepository.findByRefAndTimestampAndLimit(ref.get(), null, 0);
+            Set<String> commitIds = new HashSet<>();
```
dlamoris, line 76: not sure if you wanted the return list in order, but right now it's not

BA-MBSE: Good point, added sorting control and return limits too

dlamoris (approved):

PR #80, opened 2020-06-12
Continuous Integration

ronniscapaldi:
Docker compose, dockerfile, and updated README's


.vscode/launch.json
```diff
@@ -0,0 +1,7 @@
+{
```
dlamoris: add .vscode to the .gitignore file so it doesn't get commited

Dockerfile
```diff
@@ -1,3 +1,9 @@
...
\ No newline at end of file
+FROM gradle:jdk11
+COPY --chown=gradle:gradle . /mms
+WORKDIR /mms
+RUN ./gradlew bootJar
+
+RUN cp /mms/example/build/libs/example*.jar /app.jar
+ENTRYPOINT ["java", "--add-opens", "java.base/java.lang=ALL-UNNAMED","-jar", "/app.jar"] 
+EXPOSE 8080
+EXPOSE 5432
```
dlamoris: don't need expose 5432
also fyi, the expose 8080 is not really needed the way we're using this, but still good to have
https://we-are.bookmyshow.com/understanding-expose-in-dockerfile-266938b6a33d

README.md
```diff
@@ -8,20 +8,8 @@ The MMS SDVC is a collection of modules built on top of the Spring Framework and
...
-5.7 [Mysql Docker](https://hub.docker.com/_/mysql/)
-
-    docker run -d -e MYSQL_ROOT_PASSWORD=test1234 -e MYSQL_DATABASE=mms -p 3306:3306 mysql:5.7
-
-### Elasticsearch
-Install Elasticsearch 7.1.  If you use Docker instructions are available here: [Setting up Elasticsearch with Docker](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html)
-
-    docker run -d -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.1.1
+### Java SE Development Kit 11
+Installation instructions: [JDK-11 download](https://www.oracle.com/java/technologies/javase-jdk11-downloads.html)
```
dlamoris, line 21: so this is for a developer setup - you would still want the original instructions for postgres/mysql/elastic, since doing docker-compose is more for just running/getting the application up as quickly as possible to play with/running in ci.

I would say add a separate section before the "Developer Setup" for the docker-compose instructions (something like quick start) - for that the only thing that's needed is docker

README.md
```diff
@@ -35,8 +23,9 @@ A gradle wrapper is included in the root of this repository and can be called fr
 ### The Example Sub Project:
 1. Copy the example properties file in `example/src/main/resources/` as `application.properties`
 1. Change values for all the appropriate properties. The example file holds sane values for most properties.
-1. Setup Run and Debug configurations. The command line run command is `./gradlew bootRun`
+1. In the command line, run `docker-compose up --build` to create and start all the services from the configuration.
```
dlamoris: same comment as above - for developing/debugging running it through an ide or using gradlew is still the preferred way, just move the docker-compose instructions to a new quick start section
(the only thing needed is copying application.properties, no need to change anything in there since the test profile will override any values already

dlamoris (changes requested):

example/test-env.json
```diff
@@ -0,0 +1,10 @@
+{
+    "values": [
+        {"key": "host", "value": "http://172.18.0.4:8080"},
```
HuiJun: This needs to be determined somehow at runtime. Maybe just use an alias.

localuser/src/main/java/org/openmbee/sdvc/localuser/config/LocalUserSecurityConfig.java
```diff
@@ -55,6 +55,7 @@ public DaoAuthenticationProvider authenticationProvider() {
         try {
             userDetailsService.loadUserByUsername(adminUsername);
         } catch (UsernameNotFoundException e) {
+            userDetailsService.setPasswordEncoder(passwordEncoder());
```
HuiJun: I think this was debug code. Can you confirm that this is required, and if not, remove?

example/src/main/resources/application.properties.example
```diff
@@ -2,7 +2,7 @@
 sdvc.admin.username=test
 sdvc.admin.password=test
 
-jwt.secret=make_me_something_really_long
+jwt.secret=12345678901234567890123456789012
```
HuiJun: Revert this back.

HuiJun (changes requested):

example/test-env.json
dlamoris: have you tried just using http://mms:8080? if it's connected to the mms_default network the service name that's set in the compose file should route to the mms service

ronniscapaldi: Yes, I have tried using just mms and I get this error: getaddrinfo ENOTFOUND mms mms:8080

HuiJun (approved):

PR #81, opened 2020-06-25
MMS-97 - createOrUpdateProjects schema validation

jhang4:
Added a rejection for when input schema doesn't match one of the registered ProjectSchemas

### DIFF
--- a/crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
+++ b/crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Optional;
 import org.openmbee.sdvc.core.config.ContextHolder;
 import org.openmbee.sdvc.core.config.Privileges;
+import org.openmbee.sdvc.core.config.ProjectSchemas;
 import org.openmbee.sdvc.core.dao.ProjectDAO;
 import org.openmbee.sdvc.core.dao.ProjectIndex;
 import org.openmbee.sdvc.core.objects.ProjectsRequest;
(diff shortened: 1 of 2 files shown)


crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
```diff
@@ -112,6 +112,10 @@ public ProjectsResponse createOrUpdateProjects(
             }
 
             ProjectService ps = getProjectService(json);
+            if ((json.getProjectType()).equals("default")) {
```
dlamoris: default or null are both valid - look for the ProjectSchemas component in core and how it's used in the crud/cameo modules to register schemas - you'll need to inject that into this class in order to check, should check before the getProjectService call

you'll want to check everything still works by running the test (at the minimum crud test), and add to the crud test by trying to create a project with schema like 'invalid' and check the return code (you can modify the collection by importing it into postman, add a test, and export it again)

dlamoris (changes requested):

crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
jhang4: i added a new commit with the modified check and crud.postman_collection

example/crud.postman_collection.json
```diff
@@ -2657,6 +2657,46 @@
...
+						"type": "text/javascript"
+					}
+				}
+			],
+			"request": {
+				"method": "POST",
+				"header": [],
+				"body": {
+					"mode": "raw",
+					"raw": "{\r\n  \"projects\": [\r\n    {\r\n      \"orgId\": \"string\",\r\n      \"name\": \"string\",\r\n      \"id\": \"string\",\r\n      \"schema\":\"invalid\"\r\n    }\r\n  ]\r\n}",
```
dlamoris: since this is testing for schema, give it a valid orgId that's been created

crud/src/main/java/org/openmbee/sdvc/crud/controllers/projects/ProjectsController.java
```diff
@@ -111,7 +115,13 @@ public ProjectsResponse createOrUpdateProjects(
                 continue;
             }
 
+            if (!((schemas.getSchemas()).containsKey(json.getProjectType()))) {
```
dlamoris: this is good, we also want it to be valid if getProjectType() is null, so either check it's not null here or set it to 'default' if it's null prior

jhang4: okay I added a check in the same line

example/crud.postman_collection.json
jhang4: changed it to a valid orgId

PR #82, opened 2020-07-02
Updated for view editor compatibility

BA-MBSE:
* Added search pagination and consistent ordering
* Added /checkAuth endpoint
* Added field wildcard

Updated ElasticSearchService to intelligently choose query methods based on field mapping types

### DIFF
--- a/authenticator/src/main/java/org/openmbee/sdvc/authenticator/controllers/AuthenticationController.java
+++ b/authenticator/src/main/java/org/openmbee/sdvc/authenticator/controllers/AuthenticationController.java
@@ -5,13 +5,16 @@ import io.swagger.v3.oas.annotations.tags.Tag;
 import org.openmbee.sdvc.authenticator.security.JwtAuthenticationRequest;
 import org.openmbee.sdvc.authenticator.security.JwtAuthenticationResponse;
 import org.openmbee.sdvc.authenticator.security.JwtTokenGenerator;
+import org.openmbee.sdvc.authenticator.security.JwtTokenValidationResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
+import org.springframework.security.access.prepost.PreAuthorize;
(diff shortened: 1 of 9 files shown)


authenticator/src/main/java/org/openmbee/sdvc/authenticator/controllers/AuthenticationController.java
```diff
@@ -47,4 +50,23 @@ public JwtAuthenticationResponse createAuthenticationToken(
 
     }
 
+    @GetMapping(value = "/checkAuth")
+    @PreAuthorize("isAuthenticated()")
+    public JwtTokenValidationResponse checkAuthenticationToken() {
+        final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        if (authentication != null && authentication.getPrincipal() instanceof String) {
+            final String username = (String) authentication.getPrincipal();
+            if (username != null && username != "anonymousUser") {
```
HuiJun: Is "anonymousUser" a standard user that can be created by some dependency? Or is it a user that will be required?

dlamoris: i think that's what spring puts in when a user isn't authenticated (authenticated anonymously) 
maybe a better way would be to just check for Authentication object instance of AnonymousAuthenticationToken, 
and if it's not, just do authentication.getName() (getName already checks for userdetails/principal and calls the right things)

dlamoris (changes requested):
the check auth can be simplified (see MethodSecurityService), otherwise looks good

authenticator/src/main/java/org/openmbee/sdvc/authenticator/controllers/AuthenticationController.java
BA-MBSE: Simplified as requested

dlamoris (approved):

PR #85, opened 2020-07-30
MMS-40 - polish db initializations and population

jhang4:
Added class that checks and updates the role and privilege mappings without needing data.sql


permissions/src/main/java/org/openmbee/sdvc/permissions/config/PermissionInit.java
```diff
@@ -0,0 +1,68 @@
...
+            if (!(privIn.isPresent())) {
+                Privilege missingPriv = new Privilege();
+                missingPriv.setName(privs);
+                privRepo.saveAndFlush(missingPriv);
+            }
+        }
+
+        for (Role role : roleRepo.findAll()) {
+            Set<Privilege> pSet = new HashSet<Privilege>();
+            for (Privilege priv : privRepo.findAll()) {
```
dlamoris: can set this outside the role loop so it's not doing a findall for each role

permissions/src/main/java/org/openmbee/sdvc/permissions/config/PermissionInit.java
```diff
@@ -0,0 +1,68 @@
...
+                if (((RPmap.get(role.getName())).contains(priv.getName())) && ((role.getPrivileges() == null) || (!(role.getPrivileges().contains(priv))))) {
+                    pSet.add(priv);
+                    role.setPrivileges(pSet);
+                    roleRepo.saveAndFlush(role);
```
dlamoris: can move the save outside the priv loop
at the first instance of a priv that's not found, it's resetting the priv set to a set with only 1 priv, then it'll keep adding and saving to the set since the rest of the priv have been removed
just add all the relevant priv to the set and save at the end

dlamoris (changes requested):

permissions/src/main/java/org/openmbee/sdvc/permissions/config/PermissionInit.java
```diff
@@ -54,15 +54,18 @@ public void onApplicationEvent(final ApplicationReadyEvent event) {
...
             Set<Privilege> pSet = new HashSet<Privilege>();
-            for (Privilege priv : privRepo.findAll()) {
+            for (Privilege priv : privList) {
                 if (((RPmap.get(role.getName())).contains(priv.getName())) && ((role.getPrivileges() == null) || (!(role.getPrivileges().contains(priv))))) {
                     pSet.add(priv);
-                    role.setPrivileges(pSet);
-                    roleRepo.saveAndFlush(role);
                 }
             }
+            role.setPrivileges(pSet);
```
dlamoris, line 72: this only sets priv that are not already there and removes ones that are already there - you can either remove the entire second conditional clause, or add the priv to the existing role.getPrivileges instead of a brand new set

permissions/src/main/resources/data.sql
```diff
@@ -1,69 +0,0 @@
...
-  (2, 19),
-  (1, 1),
-  (1, 4),
-  (1, 7),
-  (1, 9),
-  (1, 13),
-  (1, 15),
-  (1, 19);
-
-INSERT INTO groups (id, created, modified, name) VALUES
```
dlamoris, line 68: also check and add the 'everyone' group

permissions/src/main/java/org/openmbee/sdvc/permissions/config/PermissionInit.java
```diff
@@ -0,0 +1,81 @@
...
+    private PrivilegeRepository privRepo;
+    @Autowired
+    private RoleRepository roleRepo;
+    @Autowired
+    private GroupRepository groupRepo;
+
+    @Override
+    public void onApplicationEvent(final ApplicationReadyEvent event) {
+
+        Map<String, List> RPmap = new LinkedHashMap<>();
```
dlamoris: you can move the map building code to the core module (under core/src/main/java/org/openmbee/sdvc/core/config), add a class or a static field in Constants and use a static block to do the inits. 
Also use the enum names instead of doing Privileges.value()[x]

permissions/src/main/java/org/openmbee/sdvc/permissions/config/PermissionInit.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.boot.context.event.ApplicationReadyEvent;
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+import javax.transaction.Transactional;
+import java.util.*;
+
+@Component
+@Transactional
+public class PermissionInit implements ApplicationListener<ApplicationReadyEvent> {
+    @Autowired
```
dlamoris: instead of having autowired on a field, prefer to use it on a setter function (autowiring on private fields is discourage)

dlamoris (changes requested):

PR #86, opened 2020-07-31
Feature/maven publish

HuiJun:


Dockerfile
```diff
@@ -1,8 +1,7 @@
 FROM gradle:jdk11
 COPY --chown=gradle:gradle . /mms
 WORKDIR /mms
-RUN ./gradlew build -x test
-RUN mkdir -p /mms/temp && find /mms/ -name '*jar' -exec cp -a -t /mms/temp {} +
+RUN ./gradlew --no-daemon build -x test
```
dlamoris, line 4: we can just go back to using a jdk image if using gradlew, otherwise just use gradle, whichever is smaller

HuiJun: true true. i do love alpine

dlamoris (approved):

PR #89, opened 2020-08-13
MMS-29 - adding useful log messages

jhang4:
added logging of requests, response body, and time duration


dlamoris:
moved to mmsri

PR #94, opened 2020-08-26
Spring Boot Actuator

ronniscapaldi:


dlamoris:
actuator incorporated into mmsri repo

PR #104, opened 2020-09-30
Adding MinIO implementation

HuiJun:


ivan-gomes:
Is this implementation intended to be MinIO specific or should it work with any S3 compatible object store?

ref: https://docs.min.io/docs/how-to-use-aws-sdk-for-java-with-minio-server.html

dlamoris:
it should be any s3 object store - so I agree it should be using the aws library and we'll want to test it against aws s3 also
the path style option should be a configurable property since aws and minio probably needs different settings
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
the properties should also be s3.x instead of minio.x

storage/src/main/java/org/openmbee/sdvc/storage/MinIOStorage.java
```diff
@@ -0,0 +1,135 @@
...
+        return location;
+    }
+
+    private String buildLocation(ElementJson element, String mimetype) {
+        int it = 1;
+        try {
+            Iterable<Result<Item>> results = getClient().listObjects(
+                ListObjectsArgs.builder().bucket(getBucket()).build());
+            for (Result<Item> result : results) {
+                it++;
```
dlamoris, line 110: generating a datetime string's probably better - saves having to list and iterate all bucket items
it's not like the numbers mean anything since it's based on all artifacts across all projects

PR #105, opened 2020-10-05
S3 Storage Implementation

HuiJun:
S3 Storage Implementation


example/src/main/resources/application.properties.example
```diff
@@ -66,3 +66,9 @@ springdoc.swagger-ui.operationsSorter=alpha
 springdoc.swagger-ui.tagsSorter=alpha
 springdoc.default-produces-media-type=application/json
 springdoc.swagger-ui.displayOperationId=true
+
+minio.endpoint=http://localhost:9000
```
dlamoris: change minio to s3, will need to add to application-test.properties too
should also add to the docker-compose file minio service

storage/src/main/java/org/openmbee/sdvc/storage/S3Storage.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.tika.mime.MimeTypes;
+import org.openmbee.sdvc.artifacts.storage.ArtifactStorage;
+import org.openmbee.sdvc.json.ElementJson;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
```
dlamoris: this can just be Component

storage/src/main/java/org/openmbee/sdvc/storage/S3Storage.java
```diff
@@ -0,0 +1,130 @@
...
+        return s3Client;
+    }
+
+    @Override
+    public byte[] get(String location, ElementJson element, String mimetype) {
+        GetObjectRequest rangeObjectRequest = new GetObjectRequest(getBucket(), location);
+        try {
+            return getClient().getObject(rangeObjectRequest).getObjectContent().readAllBytes();
+        } catch (Exception e) {
+            logger.error("Error getting object: ", e);
```
dlamoris: this should throw server error exception and not return null

storage/src/main/java/org/openmbee/sdvc/storage/S3Storage.java
```diff
@@ -0,0 +1,130 @@
...
+        ObjectMetadata om = new ObjectMetadata();
+        om.setContentType(mimetype);
+        om.setContentLength(data.length);
+
+        PutObjectRequest por = new PutObjectRequest(getBucket(), location, new ByteArrayInputStream(data), om);
+
+        try {
+            getClient().putObject(por);
+        } catch (Exception e) {
+            logger.error("Error putting object: ", e);
```
dlamoris: throw error here

storage/storage.gradle
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    api project(':artifacts')
+
+    compile group: 'com.amazonaws', name: 'aws-java-sdk-s3', version: '1.11.874'
```
dlamoris: use implementation and consistent format

storage/src/main/java/org/openmbee/sdvc/storage/S3Storage.java
```diff
@@ -0,0 +1,130 @@
...
+    private AmazonS3 getClient() {
+        if (s3Client == null) {
+            AWSCredentials credentials = new BasicAWSCredentials(ACCESS_KEY, SECRET_KEY);
+            ClientConfiguration clientConfiguration = new ClientConfiguration();
+            clientConfiguration.setSignerOverride("AWSS3V4SignerType");
+
+            s3Client = AmazonS3ClientBuilder
+                .standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(ENDPOINT, REGION))
+                .withPathStyleAccessEnabled(true)
```
dlamoris, line 64: does this work with new aws buckets or does it need to be a config?

example/artifacts.postman_collection.json
```diff
@@ -484,9 +486,9 @@
 					"script": {
 						"id": "f23976f1-afb8-495f-892a-dcbcc7360665",
 						"exec": [
-							"pm.test(\"response is jpg\", function () {\r",
-							"    pm.response.to.have.status(200);\r",
-							"    pm.response.to.be.header(\"Content-Type\", \"application/json\");\r",
+							"pm.test(\"response is jpg\", function () {",
```
dlamoris: fix test description to say json instead of jpg

example/artifacts.postman_collection.json
```diff
@@ -443,9 +445,9 @@
...
```
dlamoris: fix test description to say json instead of jpg

dlamoris (changes requested):

storage/src/main/java/org/openmbee/sdvc/storage/S3Storage.java
HuiJun: I'm iffy about this as well. S3 supports both virtual-hosted-style and path-style by default, with the only difference being that it requires a region in the endpoint for the pathStyleAccess. By default, it tries to autodetect, in case a bucketname is an invalid dns host, but I thought it would be better to actually know that it is using path style access.

dlamoris: can we do a quick test with our internal s3 for a new bucket? it's a bit unclear in the blog post with the update - it originally said any new buckets created after Sep 30, 2020 would require virtual hosted access, but that they're extending the deprecation date (it sounds like they're delaying deprecation for all buckets though)

https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

dlamoris (approved):

PR #107, opened 2020-10-08
Reconciling internal Boeing repo with open source repo

BA-MBSE:


crud/src/main/java/org/openmbee/sdvc/crud/controllers/BaseController.java
```diff
@@ -67,6 +72,14 @@ public void setObjectMapper(ObjectMapper om) {
         return om.convertValue(obj, new TypeReference<Map<String, Object>>() {});
     }
 
+    protected void findBranch(String projectId, String id){
+        getProjectType(projectId);
```
dlamoris: I think the ContextHolder context needs to be set to projectId before using branchRepository, it doesn't look like it's been set before findBranch is being called

Another option would be to check in the MethodSecurityService since that should apply more globally than calling it from each controller method

crud/src/main/java/org/openmbee/sdvc/crud/services/DefaultBranchService.java
```diff
@@ -136,11 +121,11 @@ public RefJson createBranch(String projectId, RefJson branch) {
         branch.setCreated(Formats.FORMATTER.format(now));
         branch.setDeleted(false);
         branch.setProjectId(projectId);
+        branch.setStatus("created");
 
         if (branch.getDocId() == null || branch.getDocId().isEmpty()) {
-            String uuid = UUID.randomUUID().toString();
-            branch.setDocId(uuid);
-            b.setDocId(uuid);
+            branch.setDocId(branch.getId());
```
dlamoris: the branch ids can possibly have duplicates from different projects (also master), so the docId should be unique (this is probably making an assumption about the index impl since we do use separate indexes for each project in the elastic impl)

crud/src/main/java/org/openmbee/sdvc/crud/services/NodePostHelper.java
```diff
@@ -66,20 +66,20 @@ public NodeChangeInfo processPostJson(List<ElementJson> elements, boolean overwr
             if (element.getId() == null || element.getId().isEmpty()) {
                 element.setId(UUID.randomUUID().toString());
             }
-            ElementJson elasticElement = info.getExistingElementMap().get(element.getId());
+            ElementJson indexElement = info.getExistingElementMap().get(element.getId());
             Node n = info.getExistingNodeMap().get(element.getId());
             if (n == null) {
                 added = true;
-            } else if (elasticElement == null) {
+            } else if (indexElement == null) {
```
dlamoris, line 73: can you also change the 'elastic' word in the log message to index? thx

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
```diff
@@ -79,6 +77,16 @@ public boolean createProjectDatabase(Project project) throws SQLException {
         return !created.isEmpty();
     }
 
+    public void deleteProjectDatabase(Project project) throws SQLException {
+        Connection connection = crudDataSources.getDataSource("DEFAULT").getConnection();
```
dlamoris: DEFAULT should use the ContextObject.DEFAULT_PROJECT from core
should wrap this in a try with resources/close connection per LGTM warning

dlamoris (changes requested):

crud/src/main/java/org/openmbee/sdvc/crud/controllers/BaseController.java
BA-MBSE: Yeah, I never really liked it that way either. I like the MethodSecurityService idea. Updated it to check for org/project/branch existance

crud/src/main/java/org/openmbee/sdvc/crud/services/DefaultBranchService.java
BA-MBSE: Fair point.  I moved the docId creation to the branchIndex itself to allow it to be implementation specific.

crud/src/main/java/org/openmbee/sdvc/crud/services/NodePostHelper.java
BA-MBSE: updated

rdb/src/main/java/org/openmbee/sdvc/rdb/config/DatabaseDefinitionService.java
BA-MBSE: updated

dlamoris (approved):

HuiJun (approved):

PR #113, opened 2020-10-20
Feature/readthedocs

HuiJun:


.gitignore
```diff
@@ -28,5 +28,10 @@ localhost-env.json
 /nbdist/
 /.nb-gradle/
 
+### Sphinx ###
+_build
+_static
```
dlamoris, line 33: i think only _build is generated, _static and _templates are part of source

PR #135, opened 2021-01-22
Feature/commit refactor

HuiJun:


elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -31,31 +28,39 @@ protected CommitJson newInstance() {
...
+        if (json.get(CommitJson.COMMITID).toString().isEmpty()) {
+            CommitJson commitJson = (CommitJson) json;
+            commitJson.put(CommitJson.COMMITID, UUID.randomUUID().toString());
+        }
         this.index(getIndex(), json);
     }
 
-    public Optional<CommitJson> findById(String docId) {
-        return this.findById(getIndex(), docId);
+    public void index(BaseJson json, String commitId) {
```
dlamoris: i don't think this is needed - the caller should just give the right commitJson to index and not have to know about commitId overrides

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -31,31 +28,39 @@ protected CommitJson newInstance() {
     }
 
     public void indexAll(Collection<? extends BaseJson> jsons) {
-        this.indexAll(getIndex(), jsons);
+        this.indexAll(getIndex(), associateCommits(jsons));
```
dlamoris: this should index commits 'as is' (see below for more), it shouldn't assume that indexing a collection of commit json means they're all related
also, the signature of this and index should probably change to CommitJson instead of BaseJson

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -31,31 +28,39 @@ protected CommitJson newInstance() {
...
     }
 
     public void index(BaseJson json) {
+        if (json.get(CommitJson.COMMITID).toString().isEmpty()) {
```
dlamoris: I think what should happen here (and in indexAll) is the breaking up of a single commit json into several if its arrays exceeds some threshold - so the caller can give a huge object, but because of the elasticsearch limits, only the elastic impl knows to break it up (similar to how the get knows to munge commit objects together and can return a huge object to the caller, the caller doesn't need to know about munging)

At the same time, if the caller already broke up the commits, this would still work as expected

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -31,31 +28,39 @@ protected CommitJson newInstance() {
...
     }
 
-    public void deleteById(String docId) {
-        this.deleteById(getIndex(), docId);
+    public List<CommitJson> findAllById(Set<String> commitIds) {
+        return getFullCommits(commitIds);
+    }
+
+    public void deleteById(String commitId) {
+        this.deleteById(getIndex(), commitId);
```
dlamoris: this should delete all docs that have commitId, right now it treats commitId as the docId

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -31,31 +28,39 @@ protected CommitJson newInstance() {
...
     }
 
     public void deleteAll(Collection<? extends BaseJson> jsons) {
         this.deleteAll(getIndex(), jsons);
     }
 
-    public boolean existsById(String docId) {
-        return this.existsById(getIndex(), docId);
+    public boolean existsById(String commitId) {
+        return this.existsById(getIndex(), commitId);
```
dlamoris: same as delete ?

crud/src/main/java/org/openmbee/mms/crud/services/DefaultNodeService.java
```diff
@@ -194,6 +194,11 @@ public ElementsResponse createOrUpdate(String projectId, String refId, ElementsR
             .processPostJson(req.getElements(), overwriteJson,
                 createCommit(user, refId, projectId, req), this);
 
+        String commitId = params.get("commitId");
+        if (commitId != null && !commitId.isEmpty()) {
+            info.getCommitJson().setCommitId(commitId);
```
dlamoris: this should happen during the createCommit in the previous call, otherwise the elements getting updated will have the wrong commitId

crud/src/main/java/org/openmbee/mms/crud/services/NodeOperation.java
```diff
@@ -66,6 +66,7 @@ public void setBranchRepository(BranchDAO branchRepository) {
 
     public void initCommitJson(CommitJson cmjs, Instant now) {
         cmjs.setId(UUID.randomUUID().toString());
+        cmjs.setCommitId(UUID.randomUUID().toString());
```
dlamoris: this should check if commitId is not set first, if it's not set, it should just be same as the id so it's still consistent with any existing paradigm

crud/src/main/java/org/openmbee/mms/crud/services/NodeOperation.java
```diff
@@ -130,7 +131,7 @@ public void processElementAdded(ElementJson e, Node n, NodeChangeInfo info) {
         cmjs.getAdded().add(newObj);
 
         n.setNodeId(e.getId());
-        n.setInitialCommit(e.getDocId());
+        n.setInitialCommit(e.getCommitId());
```
dlamoris, line 138: this was actually a misnomer and the initial commit was referring to the initial element's doc id - i don't think that's relevant anymore though, just fyi (i think it was because of the initial commit object issue before too)

dlamoris (changes requested):
initial comments

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -18,29 +18,74 @@
...
+                            break;
+                        case "deleted":
+                            currentCommitCopy.getDeleted().add(action);
+                            break;
+                    }
+                }
+                broken.add(currentCommitCopy);
+
+            }
+            this.indexAll(broken);
```
HuiJun, line 76: Just fyi, this adds a property called "action" into the commit objects. We could remove it if it's a problem.

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
HuiJun: Ok, removed

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
HuiJun: removed

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
HuiJun: Done, but it might be done in a different way

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
HuiJun: Done

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
HuiJun: done

crud/src/main/java/org/openmbee/mms/crud/services/DefaultNodeService.java
HuiJun: done

crud/src/main/java/org/openmbee/mms/crud/services/NodeOperation.java
HuiJun: done

crud/src/main/java/org/openmbee/mms/crud/services/DefaultNodeService.java
```diff
@@ -220,16 +221,23 @@ protected void commitChanges(NodeChangeInfo info) {
...
-                Commit commit = new Commit();
-                commit.setBranchId(cmjs.getRefId());
-                commit.setCommitType(CommitType.COMMIT);
-                commit.setCreator(cmjs.getCreator());
-                commit.setDocId(cmjs.getId());
-                commit.setTimestamp(now);
-                commit.setComment(cmjs.getComment());
-                
-                this.commitRepository.save(commit);
+                Optional<Commit> existing = this.commitRepository.findByCommitId(cmjs.getCommitId());
```
dlamoris: cmjs.getId()

crud/src/main/java/org/openmbee/mms/crud/services/DefaultNodeService.java
```diff
@@ -294,13 +302,18 @@ public ElementsResponse delete(String projectId, String refId, ElementsRequest r
...
+        ElementsRequest req, String commitId) {
         CommitJson cmjs = new CommitJson();
         cmjs.setCreator(creator);
         cmjs.setComment(req.getComment());
         cmjs.setSource(req.getSource());
         cmjs.setRefId(refId);
         cmjs.setProjectId(projectId);
+
+        if (commitId != null && !commitId.isEmpty()) {
+            cmjs.setId(commitId);
```
dlamoris, line 314: there's somewhere else (initCommit)? that needs to be updated to check for existing id and generate docId

crud/src/main/java/org/openmbee/mms/crud/services/NodeOperation.java
```diff
@@ -163,7 +163,7 @@ private void processElementAddedOrUpdated(ElementJson e, Node n, NodeChangeInfo
         }
 
         n.setDocId(e.getDocId());
-        n.setLastCommit(cmjs.getId());
+        n.setLastCommit(cmjs.getCommitId());
```
dlamoris: getId()

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -35,27 +36,70 @@ public void indexAll(Collection<? extends BaseJson> jsons) {
...
+                        case "added":
+                            currentCommitCopy.getAdded().add(action);
+                            break;
+                        case "updated":
+                            currentCommitCopy.getUpdated().add(action);
+                            break;
+                        case "deleted":
+                            currentCommitCopy.getDeleted().add(action);
+                            break;
+                    }
```
dlamoris, line 71: can remove action here?

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -127,4 +170,107 @@ protected String getIndex() {
     public CommitJson update(CommitJson commitJson) {
         return this.update(getIndex(), commitJson);
     }
+
+    private List<CommitJson> getDocs(String commitId) {
+        try {
+            QueryBuilder commitQuery = QueryBuilders.boolQuery()
+                .should(QueryBuilders.termQuery(CommitJson.ID, commitId))
```
dlamoris: this is better as a filter instead of should

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -35,27 +36,70 @@ public void indexAll(Collection<? extends BaseJson> jsons) {
...
     public void deleteAll(Collection<? extends BaseJson> jsons) {
         this.deleteAll(getIndex(), jsons);
     }
 
-    public boolean existsById(String docId) {
-        return this.existsById(getIndex(), docId);
+    public boolean existsById(String commitId) {
+        List<CommitJson> docs = getDocs(commitId);
+        List<Boolean> result = new ArrayList<>();
+        docs.forEach(commit -> {
```
dlamoris: isn't it enough to check the list isn't empty? the fact it returns the object means it exists?

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -21,11 +18,15 @@
...
 import org.openmbee.mms.json.BaseJson;
 import org.openmbee.mms.json.CommitJson;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 @Component
 public class CommitElasticDAOImpl extends BaseElasticDAOImpl<CommitJson> implements CommitIndexDAO {
 
+    @Value("${elasticsearch.limit.index}")
+    int indexLimit;
```
dlamoris: should we be using the same limit for indexing vs splitting up commit? the commit can be a lot bigger than whatever we set the index limit to (currently used in bulk processor) (this is already injected in the BaseDao as bulkLimit)

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -127,4 +170,107 @@ protected String getIndex() {
...
+                partial.setComment("");
+            }
+            if (partial.getId() == null) {
+                partial.setId("");
+            }
+            if (partial.getId() == null) {
+                partial.setId("");
+            }
+
+            if (partial.getAdded().isEmpty() && raw.getAdded() != null) {
```
dlamoris: I don't think the first check should be here? this would only add addition data if partial is empty but we want to add them regardless

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -127,4 +170,107 @@ protected String getIndex() {
...
+            if (partial.getSource() == null) {
+                partial.setSource("");
+            }
+            if (partial.getComment() == null) {
+                partial.setComment("");
+            }
+            if (partial.getId() == null) {
+                partial.setId("");
+            }
+            if (partial.getId() == null) {
```
dlamoris: duplicate

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -127,4 +170,107 @@ protected String getIndex() {
...
+            if (partial.getDeleted().isEmpty() && raw.getDeleted() != null) {
+                partial.getDeleted().addAll(raw.getDeleted());
+            }
+            if (partial.getSource().isEmpty() && raw.getSource() != null) {
+                partial.setSource(raw.getSource());
+            }
+            if (partial.getComment().isEmpty() && raw.getComment() != null) {
+                partial.setComment(raw.getComment());
+            }
+            if (partial.getId().isEmpty() && raw.getCommitId() != null) {
```
dlamoris: duplicate

crud/src/main/java/org/openmbee/mms/crud/controllers/elements/ElementsController.java
```diff
@@ -113,6 +126,54 @@ public ElementsResponse createOrUpdateElements(
         throw new BadRequestException(response.addMessage("Empty"));
     }
 
+    @PostMapping(value = "/stream", consumes = MediaType.APPLICATION_JSON_VALUE)
```
dlamoris, line 129: can comment this function out for now so it doesn't show up in generated swagger

dlamoris (changes requested):

elastic/src/main/java/org/openmbee/mms/elastic/CommitElasticDAOImpl.java
```diff
@@ -93,26 +137,25 @@ private QueryBuilder getCommitHistoryQuery(String id, Set<String> commitIds) {
...
+                CommitJson ob = new CommitJson();
+                ob.putAll(hit.getSourceAsMap());
+                if (!rawCommits.containsKey(ob.getId())) {
+                    rawCommits.put(ob.getId(), new ArrayList<>());
+                }
+                rawCommits.get(ob.getId()).add(ob); // gets "_source"
             }
-            return commits;
+            ArrayList<CommitJson> result = new ArrayList<>();
+            for (String key : rawCommits.keySet()) {
```
dlamoris: does using keySet retain ordering?
also because of the query, the commit objects returned wouldn't be the entire object if it was split - we used to just return the commit id and some metadata for element history, maybe we can just do that again

example/crud.postman_collection.json
```diff
@@ -2467,54 +2467,13 @@
 			},
 			"response": []
 		},
-		{
```
dlamoris: revert deletion

elastic/src/main/resources/elastic_mappings/commit.json
```diff
@@ -59,7 +59,7 @@
             "_projectId": {
                 "type": "keyword"
             },
-            "_docId": {
```
dlamoris: revert

json/src/main/java/org/openmbee/mms/json/CommitJson.java
```diff
@@ -17,6 +19,22 @@
     public static final String UPDATED = "updated";
     public static final String SOURCE = "source";
 
+    public static CommitJson copy(CommitJson original) {
+        CommitJson copy = new CommitJson();
```
dlamoris: copy comment

dlamoris (approved):

PR #137, opened 2021-02-02
Synchronizing changes from Boeing to open source

BA-MBSE:
Update ElementsResponse to include commit id when committing changes
Branch id input sanitization
Add option to disable TWC auth delegation & host parsing cleanup
Added TokenService to allow use of tokens in other contexts
Allow password setting for admins (user conversion from external to internal)
Handle setting of null passwords (user conversion from internal to external)

### DIFF
--- a/core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
+++ b/core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
@@ -1,0 +1,13 @@
+package org.openmbee.mms.core.objects;
+
+public class ElementsCommitResponse extends ElementsResponse {
+    private String commitId;
+
+    public String getCommitId() {
+        return commitId;
+    }
+
+    public void setCommitId(String commitId) {
+        this.commitId = commitId;
+    }
+}
--- a/core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
+++ b/core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
@@ -1,6 +1,10 @@
 package org.openmbee.mms.core.objects;
 
+import io.swagger.v3.oas.annotations.media.Schema;
+
 public class ElementsCommitResponse extends ElementsResponse {
+
+    @Schema(nullable = true)
     private String commitId;
 
     public String getCommitId() {
--- a/rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
+++ b/rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
@@ -9,6 +9,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.transaction.PlatformTransactionManager;
 
+import java.util.regex.Pattern;
+
 public abstract class BaseDAOImpl {
 
     private CrudDataSources crudDataSources;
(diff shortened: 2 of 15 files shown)


rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
```diff
@@ -36,6 +38,10 @@ public JdbcTemplate getConn() {
 
     public String getSuffix() {
         String refId = ContextHolder.getContext().getBranchId();
-        return refId.equals(ContextObject.MASTER_BRANCH) ? "" : refId.toLowerCase();
+        if(Pattern.matches("^[a-zA-Z0-9-_]*$", refId)) {
```
HuiJun: getSuffix() is used in a number of places multiple times in our DAOs. Recompiling the pattern every time could potentially have an effect on performance, albeit, probably not significantly.

twc/src/main/java/org/openmbee/mms/twc/config/TwcConfig.java
```diff
@@ -39,11 +51,21 @@ public TwcAuthenticationProvider getAuthNProvider(String associatedTWC) {
...
-            if (twc.hasKnownName(twcUrl)) {
+            if (twc.hasKnownName(host)) {
                 return twc;
             }
         }
         return null;
     }
+
+    private static String stripHost(String url) {
+        Pattern pattern = Pattern.compile("(https?://)?([\\w-\\.]*)(:\\d+)?");
```
HuiJun, line 64: Should check this and possibly move it to a static final.

HuiJun (changes requested):

core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
```diff
@@ -0,0 +1,13 @@
+package org.openmbee.mms.core.objects;
+
+public class ElementsCommitResponse extends ElementsResponse {
+    private String commitId;
```
dlamoris, line 8: add @Schema(nullable = true) for openapi

rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
dlamoris: there's a BRANCH_ID_VALID_PATTERN in the BranchesController - may be better if that gets moved to core Formats?

HuiJun: i think so

rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
```diff
@@ -38,7 +40,7 @@ public JdbcTemplate getConn() {
 
     public String getSuffix() {
         String refId = ContextHolder.getContext().getBranchId();
-        if(Pattern.matches("^[a-zA-Z0-9-_]*$", refId)) { 
+        if(Pattern.matches(BRANCH_ID_VALID_PATTERN, refId)) {
```
HuiJun: From what I understand Pattern.matches(String, String) is a convenience method that does Pattern.matches(Pattern.compile(String), String) for regex patterns that are used only once. The expensive bit is Pattern.compile, which should instead be defined as static final so that it is only compiled once at initialization.

HuiJun (changes requested):

rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
BA-MBSE: Changed to a static Pattern instead

core/src/main/java/org/openmbee/mms/core/objects/ElementsCommitResponse.java
BA-MBSE: added

rdb/src/main/java/org/openmbee/mms/rdb/repositories/BaseDAOImpl.java
BA-MBSE: updated to use a common Pattern

dlamoris (approved):

HuiJun (approved):

PR #155, opened 2021-02-20
add checks for 500 and 409 errors, use sha-1 as suffix for branches with id > 50

dlamoris:


HuiJun (approved):

HuiJun (approved):
Small note, maybe we should name getSuffix something more appropriate now.

PR #156, opened 2021-02-24
Element Level Permissions example Schema

BA-MBSE:
Adding an example implementation for element level permissions - thi is per element only by group and does not handle any kind of inheritance


HuiJun (approved):

crud/src/main/java/org/openmbee/mms/crud/services/NodeDeleteHelper.java
```diff
@@ -23,6 +25,10 @@ public NodeChangeInfo processDeleteJson(List<ElementJson> elements, CommitJson c
             }
             Node node = info.getExistingNodeMap().get(nodeId);
             Map<String, Object> indexElement = info.getExistingElementMap().get(nodeId);
+            if(! validateOperation(Operation.REMOVE, null, indexElement, false)) {
+                info.addRejection(nodeId, new Rejection(nodeId, 406, "Invalid operation"));
```
HuiJun, line 29: Maybe we should throw the actual validation error, e.g. 401, 403. In the future, we may want to process other validation besides permissions checking.

HuiJun (approved):

PR #164, opened 2021-05-06
Add more logging when commit error happens

HuiJun:


crud/src/main/java/org/openmbee/mms/crud/services/DefaultNodeService.java
```diff
@@ -234,7 +236,12 @@ public void commitChanges(NodeChangeInfo info) {
                     });
                 this.nodeRepository.saveAll(new ArrayList<>(nodes.values()));
             } catch (Exception e) {
-                logger.error("commitChanges error: {}", e.getMessage());
+                StringWriter sw = new StringWriter();
+                PrintWriter pw = new PrintWriter(sw);
+                e.printStackTrace(pw);
+                String sStackTrace = sw.toString();
+
+                logger.error("Error in commitChanges: {}", sStackTrace);
```
dlamoris: can do logger.error("Error in commitChanges:", e) so don't need to use printwriter

dlamoris (approved):

PR #173, opened 2021-09-20
Enhance Users controller and LDAP logic

### LINKED ISSUE #172: Non-admins should be able to view users
In order to facilitate users being able to administer their own projects, users should be able to lookup other users without having admin privileges.

Enquier:
This is a proposed fix for #172 which allows non-admins to lookup users. It also adds a query parameter which to allow search for a specific username.


ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
```diff
@@ -178,4 +189,28 @@ public BaseLdapPathContextSource contextSource() {
...
+        }
+        if (!saveUser.getFirstName().equals(userData.getStringAttribute(userAttributesFirstName))) {
+            saveUser.setFirstName(userData.getStringAttribute(userAttributesFirstName));
+        }
+        if (!saveUser.getLastName().equals(userData.getStringAttribute(userAttributesLastName))) {
+            saveUser.setLastName(userData.getStringAttribute(userAttributesLastName));
+        }
+
+        saveUser.setEnabled(true);
+        saveUser.setAdmin(false);
```
HuiJun: These will need to be considered. We don't want to modify an admin user and have this logic convert them to a regular user and we don't want to enable a disabled user on modification either.

build.gradle
```diff
@@ -157,4 +157,37 @@ subprojects {
             sign publishing.publications.mavenJava
         }
     }
-}
\ No newline at end of file
+
+}
+
```
HuiJun: Why is this?

HuiJun (changes requested):

ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
Enquier: Fixed. Moved over to the create side only

build.gradle
Enquier: Adds ability to run MMS as a server from gradle directly. For rapid development without generating a new docker container for each run

HuiJun: MMS already has the ability to run from gradle directly. It's called ./gradlew bootRun.

HuiJun (changes requested):

HuiJun (approved):

PR #176, opened 2021-10-07
Release/4.0.5

HuiJun:


ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
```diff
@@ -178,4 +189,28 @@ public BaseLdapPathContextSource contextSource() {
         return contextSource;
     }
 
+    private User saveLdapUser(DirContextOperations userData, User saveUser) {
```
dlamoris, line 192: can any of the the getXXX from saveUser be null? need to check for npe

dlamoris (changes requested):

ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
```diff
@@ -178,4 +189,28 @@ public BaseLdapPathContextSource contextSource() {
...
+            saveUser.setFirstName(userData.getStringAttribute(userAttributesFirstName));
+        }
+        if (!saveUser.getLastName().equals(userData.getStringAttribute(userAttributesLastName))) {
+            saveUser.setLastName(userData.getStringAttribute(userAttributesLastName));
+        }
+
+        return saveUser;
+    }
+
+    private User createLdapUser(DirContextOperations userData) {
```
dlamoris, line 206: should also try to init first and last name on create here? otherwise it'll be filled in if they log in again after 24 hours

dlamoris (approved):

PR #187, opened 2022-05-17
Feature/update jdk

HuiJun:


example/example.gradle
```diff
@@ -45,4 +45,7 @@ configurations {
         it.outgoing.artifact(bootJar)
     }
 }
+bootJar {
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
dlamoris, line 49: do you remember which warning was triggered, or what was being duplicated?

dlamoris (approved):

example/example.gradle
HuiJun: `Entry BOOT-INF/lib/jna-5.5.0.jar is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/7.4.2/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.`

PR #192, opened 2022-06-15
Feature/group permissions

Enquier:
This PR adds more control over the management and creation of groups, it also refactors the UserDetails service to provide a more abstract users api


HuiJun:
Firstly, I think we should discuss why this is needed at all. I think originally, we wanted as much abstraction with 'groups' as possible and let whichever authentication implementation handle it however it wanted. Secondly, would such fine grained authorization even be necessary, as we can just let an 'admin' user do whatever they wanted. It just seems like a lot of overhead just to allow fine grained control of an object we don't even want to manage in the first place. Lastly, a local and remote user could easily be differentiated by whether a password is set or not, so mangling the data structure doesn't seem to be necessary there either. I think a better approach might be to just remove the 'syncing' of user and groups altogether. Anyways, further discussion is needed before more work is done here.

permissions/src/main/java/org/openmbee/mms/permissions/config/PermissionInit.java
```diff
@@ -76,10 +98,30 @@ public void onApplicationEvent(final ApplicationReadyEvent event) {
...
         Optional<Group> evGroupIn = groupRepo.findByName("everyone");
-        if (!(evGroupIn.isPresent())) {
-            Group evGroup = new Group();
+        Group evGroup;
+        if (evGroupIn.isEmpty()) {
+            Optional<Role> evRole = roleRepo.findByName(AuthorizationConstants.READER);
+            evGroup = new Group();
             evGroup.setName("everyone");
+            evGroup.setType(Group.VALID_GROUP_TYPES.LOCAL);
+            evGroup.getUsers().addAll(userRepo.findAll());
```
dlamoris: i would rather not have to maintain all users to the everyone group, i think everyone can be injected into the core security check like org.openmbee.mms.core.utils.AuthenticationUtils

data/src/main/java/org/openmbee/mms/data/domains/global/Group.java
```diff
@@ -1,40 +1,110 @@
...
+@Table(name = "groups", uniqueConstraints = @UniqueConstraint(columnNames = "name"))
 public class Group extends Base {
 
 
     @Column(unique = true)
     private String name;
 
-    @ManyToMany(mappedBy = "groups")
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(name = "users_groups", joinColumns = @JoinColumn(name = "users_id", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "groups_id", referencedColumnName = "id"))
```
dlamoris: the column names should match current (user_id and group_id) to minimize db migration

data/src/main/java/org/openmbee/mms/data/domains/global/Group.java
```diff
@@ -1,40 +1,110 @@
...
     private Collection<User> users;
 
-    public Group() {}
+    @JsonIgnore
+    private VALID_GROUP_TYPES type;
```
dlamoris, line 24: why is both type and typeString needed?

data/src/main/java/org/openmbee/mms/data/domains/global/User.java
```diff
@@ -39,7 +47,7 @@ public class User extends Base {
 
     @JsonIgnore
     @ManyToMany(fetch = FetchType.EAGER)
-    @JoinTable(name = "users_groups", joinColumns = @JoinColumn(name = "user_id", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "group_id", referencedColumnName = "id"))
+    @JoinTable(name = "users_groups", joinColumns = @JoinColumn(name = "users_id", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "groups_id", referencedColumnName = "id"))
```
dlamoris: avoid changing existing column names

data/src/main/java/org/openmbee/mms/data/domains/global/User.java
```diff
@@ -20,6 +20,14 @@ public class User extends Base {
     private String lastName;
     private boolean admin;
 
+    public enum VALID_USER_TYPES {LOCAL, REMOTE}
```
dlamoris, line 27: we currently distinguish local vs remote user by empty password, which should be sufficient

docker-compose.yml
```diff
@@ -27,15 +27,15 @@ services:
...
-    environment:
-        - "SPRING_PROFILES_ACTIVE=test"
-    depends_on:
-        - postgres
-        - elasticsearch
-        - minio
-    ports: 
-        - 8080:8080
\ No newline at end of file
+  # mms:
```
dlamoris: ?
this is needed for the ci tests to work

ldap/src/main/java/org/openmbee/mms/ldap/security/LdapAuthoritiesConfig.java
```diff
@@ -159,19 +115,20 @@ public Collection<? extends GrantedAuthority> getGrantedAuthorities(
                         groupRoleAttribute);
 
                 Set<Group> addGroups = new HashSet<>();
+                Optional<Group> evGroup = getGroupRepo().findByName(AuthorizationConstants.EVERYONE);
```
dlamoris: I think the addGroups wasn't supposed to be here - for ldap/remote users we don't want to sync the groups relationships, only to return the right authorities based on the groups mms already has, if a user is added to a local group doing this will reset those relationships
lines 117 to 125 can be deleted
instead, it can find any local group relationship and add them to memberGroups, that's where the authorities are populated from

ldap/src/main/java/org/openmbee/mms/ldap/security/LdapAuthoritiesConfig.java
```diff
@@ -189,7 +146,35 @@ public BaseLdapPathContextSource contextSource() {
...
+        user.setEnabled(true);
+        user.setAdmin(req.isAdmin());
+        user.setType(req.getType());
+        return saveUser(user);
+    }
+
+    @Override
+    public User saveUser(User user) {
+
+        Optional<Group> evGroup = getGroupRepo().findByName(AuthorizationConstants.EVERYONE);
```
dlamoris: i don't think adding everyone relationships is needed, see other comment about doing it in org.openmbee.mms.core.utils.AuthenticationUtils

rdb/src/main/java/org/openmbee/mms/rdb/repositories/GroupGroupPermRepository.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Set;
+
+@Repository
+public interface GroupGroupPermRepository extends JpaRepository<GroupGroupPerm, Long> {
+
+    List<GroupGroupPerm> findAllByGroup(Group group);
+
+    List<GroupGroupPerm> findAllByGroup_Name(String group);
+
+    Optional<GroupGroupPerm> findByGroupAndGroup(Group b, Group group);
```
dlamoris: i don't think this would work, should be findByGroupAndGroupPerm

settings.gradle
```diff
@@ -28,3 +28,7 @@ buildFiles.each { File buildFile ->
         project.buildFileName = buildFile.name
     }
 }
+include 'users'
```
dlamoris: shouldn't be needed?

dlamoris (changes requested):
first pass

data/src/main/java/org/openmbee/mms/data/domains/global/User.java
Enquier: Closing this as the preference stated at the Meeting was to keep this definition

docker-compose.yml
Enquier: Accidental from my own testing, I will fix this

ldap/src/main/java/org/openmbee/mms/ldap/security/LdapAuthoritiesConfig.java
Enquier: In response to the discussion I have removed the syncing of group membership from the LDAP Authorities config

rdb/src/main/java/org/openmbee/mms/rdb/repositories/GroupGroupPermRepository.java
Enquier: Removed

settings.gradle
Enquier: fixed

ldap/src/main/java/org/openmbee/mms/ldap/security/LdapAuthoritiesConfig.java
Enquier: I think that's what was done before. I think my primary thought there was having an "everyone" group to allow assigning permissions to "everyone", I'm not sure if that is the exact same use case as "public" but I remember it wasn't working in our MMS4 tests. We could probably do both but not track the "everyone" membership like we do with remote groups

users/src/main/java/org/openmbee/mms/users/controller/UsersController.java
```diff
@@ -49,25 +51,37 @@ public UsersResponse getUsers(@RequestParam(required = false) String user) {
...
+            users.add(usersDetailsService.loadUserByUsername(user).getUser());
         } else {
-            users = userDetailsService.getUsers();
+            users = usersDetailsService.getUsers();
         }
         res.setUsers(users);
         return res;
     }
 
+    @GetMapping(value = "/whoami")
```
HuiJun, line 75: I think this is probably the same information given back on users?user=USERID

data/src/main/java/org/openmbee/mms/data/domains/global/Group.java
Enquier: fixed

data/src/main/java/org/openmbee/mms/data/domains/global/User.java
Enquier: fixed

ldap/src/main/java/org/openmbee/mms/ldap/security/LdapAuthoritiesConfig.java
Enquier: I went through and changed the Everyone group to "remote" and removed all of the user sync stuff. I then re-added the simpleAuthorityGrant for Everyone

users/src/main/java/org/openmbee/mms/users/controller/UsersController.java
Enquier: So this is a little different in that it is getting the userId from the token, i.e. it is telling you who is logged in, something that is very useful for VE and other webapps

data/src/main/java/org/openmbee/mms/data/domains/global/Group.java
Enquier: type is an enum, typeString lets you set where the source is, basically if typeString isn't "local" the group is REMOTE but the individual auth systems can set type to be "ldap" or "twc"

permissions/src/main/java/org/openmbee/mms/permissions/config/PermissionInit.java
Enquier: Resolved. Removed Everyone Group

PR #195, opened 2022-08-07
fix: support users and groups in distinct DN bases

### LINKED ISSUE #194: LDAP - Issue with users and groups in different DN bases
Consider a possible and common LDAP structure where users are in one organizational unit (ou) and groups are in another, e.g.
* users in `ou=users,dc=example,dc=com`, ex. `cn=billy,ou=users,dc=example,dc=com`
* groups in `ou=groups,dc=example,dc=com`, ex. `cn=sysadmins,ou=groups,dc=example,dc=com`

As implemented MMS cannot support this LDAP structure because a common `ldap.provider.base` configuration is used for both users and groups, implicitly via `ldap.provider.url=ldaps://dir.example.com/${ldap.provider.base}` and explicitly via https://github.com/Open-MBEE/mms/blob/1d4e8b279a33980bc3d0e0ef22339c281516bb8c/ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java#L142

A solution would be to independently and explicitly configure search bases for users and groups, as is already supported by the vanilla Spring Boot `LdapAuthenticationProvider` being overridden.

ivan-gomes:
resolves #194 

The important changes are:
* use of `userData.getDn()` instead of `userAttributesUsername + '=' + username` to get the entire DN, e.g. if `ldap.user.dn.pattern=uid={0},ou=users` this will resolve to `uid=billy,ou=users` instead of just `uid=billy`
* provide `groupSearchBase` for the group LDAP search

This change should be fully backwards compatible, i.e. should not require any configuration change for currently functioning deployments. It will add support for LDAP structures (currently incompatible) as described in #194 using a configuration like the following:

```
ldap.provider.base=dc=example,dc=com
ldap.provider.url=ldap://ldap.example.com
ldap.user.dn.pattern=uid={0},ou=users
ldap.group.search.base=ou=groups
```

### DIFF
--- a/ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
+++ b/ldap/src/main/java/org/openmbee/mms/ldap/LdapSecurityConfig.java
@@ -113,7 +113,7 @@ public class LdapSecurityConfig {
 
         /*
           Specificity here : we don't get the Role by reading the members of available groups (which is implemented by
-          default in Spring security LDAP), but we retrieve the groups from the field memberOf of the user.
+          default in Spring security LDAP), but we retrieve the groups the user belongs to.
          */
         class CustomLdapAuthoritiesPopulator implements LdapAuthoritiesPopulator {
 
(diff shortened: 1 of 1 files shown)


HuiJun (approved):
I'm an old man, and I like the faster performing 'for', but everything else seems good to me.

PR #200, opened 2022-09-14
return element json if artifact mime type not found

dlamoris:
because of content negotiation, the get element endpoint goes to artifact controller if accept header is anything other than application/json or empty, browsers put stuff incompatible stuff in there, this is so a browser link returns something useful besides artifact not found


HuiJun (approved):
LGTM

PR #209, opened 2022-10-07
feat(groups): allow all printable ASCII characters in group name

### LINKED ISSUE #208: Group name is too restricted
Currently only word characters, e.g. letters in the alphabet and numbers, and hyphens are allowed in a group's name.

https://github.com/Open-MBEE/mms/blob/3b4c6c53e5f8ea04e93be7e52ef2c87cec0e8141/groups/src/main/java/org/openmbee/mms/groups/services/GroupValidationService.java#L16

Given that MMS groups are intended to integrate with pre-existing groups defined externally in an enterprise, e.g. LDAP groups, this is overly prescriptive as often one does not have control over the enterprise's naming conventions. A practical example of this is when there are spaces in an existing group name, e.g. `Internal Users`, which is currently rejected by MMS.

My proposal is to widen the limitation to all printable characters - regex `[ -~]`, i.e. all characters between space and tilde.

ivan-gomes:
closes #208 

Operative change:
https://github.com/Open-MBEE/mms/blob/501e5d28c53a8658d100179ea3de7c42ba29a8ff/groups/src/main/java/org/openmbee/mms/groups/services/GroupValidationService.java#L16

### DIFF
--- a/groups/src/main/java/org/openmbee/mms/groups/constants/GroupConstants.java
+++ b/groups/src/main/java/org/openmbee/mms/groups/constants/GroupConstants.java
@@ -6,7 +6,7 @@ public class GroupConstants {
     public static final String GROUP_NOT_EMPTY = "Group is not empty";
     public static final String GROUP_NOT_FOUND = "Group not found";
     public static final String INVALID_ACTION = "Invalid action";
-    public static final String INVALID_GROUP_NAME= "Invalid group name";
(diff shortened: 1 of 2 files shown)


HuiJun (approved):

HuiJun:
Wouldn't this only match the beginning of the string btw?

PR #235, opened 2023-03-17
Persistence refactor

BA-MBSE:
This PR decouples the federated (ES + PG) persistence layer from the Controller and Service layers, and replaces it with a unified persistence interface layer.  Provided is a rehost of the ES + PG persistence layer as an implementation of the unified persistence layer (federatedpersistence).

Decoupled view controller from cameo module, added view module.

Added msosa module as schema for 2021x + MagicDraw Schema

Added oauth module


BA-MBSE:
@HuiJun How do you want to handle all of the "This log entry depends on a user defined value." errors?

HuiJun:
@BA-MBSE I think the play here is to ignore them for now, with a note to fix before the next release.

dlamoris:
branch was pushed to main repo, see #259

PR #252, opened 2023-12-05
Release/4.0.19

HuiJun:


dlamoris (changes requested):
check the storage/artifact aws lib version to see if it works for eks service account

also readthedocs needs a yaml file in root now to build, https://blog.readthedocs.com/migrate-configuration-v2/

also take out the lgtm badge and fix the circleci link in readme

HuiJun:
The minimum version for AWS SDK for service accounts, I believe is `1.11.704` and we are currently on `1.11.874`, so we should be good there.

.circleci/config.yml
```diff
@@ -69,21 +80,30 @@ workflows:
     version: 2
     build-test-deploy:
       jobs:
-        - build_and_test:
+        - build_and_test
```
dlamoris: i believe this will also require the filter tags to be present or else the release won't run, had something similar happen in the flexo repos

.circleci/config.yml
```diff
@@ -69,21 +80,30 @@ workflows:
...
+              context:
+                  - SonarCloud
         - deploy_snapshot:
               requires:
-                  - build_and_test
+                  - scan
               filters:
                   branches:
                       only: /((release|hotfix|support)/[0-9.]+(-(a|b|rc)[0-9]+)?|develop)/
+              context:
```
dlamoris: we dont' actually need dockerhub here since we're not pushing to docker, may want it in mmsri

dlamoris (approved):

PR #256, opened 2024-01-03
Propose Elastic Authentication capabilities

Enquier:
This is a small update which should detect if you are using authentication for elastic connection and use it.


HuiJun (approved):
We just need to fix the tests failing.

HuiJun (changes requested):
So you either need to add a null check, to handle cases when the property is not set, or uncomment the property below to avoid the NPE on startup since you're calling isEmpty() on a null.

dlamoris (approved):

PR #258, opened 2024-02-07
Reconcile OpenMBEE and Boeing

HuiJun:
This is a combined PR for resyncing Open-MBEE build with Boeing's code base.


dlamoris:
this is superceded by #259 which does not included element level permissions feature

PR #259, opened 2024-02-07
Testing BA-MBSE Persistence Refactor Merge

HuiJun:
deconflicted branch of https://github.com/Open-MBEE/exec-mms/pull/235

PR #260, opened 2024-02-13
Fix/persistence refactor fixes

dlamoris:
fixes for comments in https://github.com/Open-MBEE/exec-mms/pull/259


dlamoris:
still need to fix cameo tests to take out childViews

HuiJun (approved):

Issue #261, opened 2024-02-13
Missing check for time of element deletion while getting all elements under commitId


borozcod:
## What is the issue?
There is a case when you are trying to get all the elements from a ref at a specific commit. If the element was deleted after the commit, it will also be missing from the elements request at a earlier commit.

## How to reproduce?
1. Create elements x, y, z
2. Update y, save commit id
3. Delete element x
4. Get elements at step 2 via `/projects/aa/refs/{refId}/elements?commitId={step 2 commit id}`

I believe the check should be happening here

https://github.com/Open-MBEE/exec-mms/blob/adc1e8ca11babc6793a9370c1166cf220e6eb926/crud/src/main/java/org/openmbee/mms/crud/services/NodeGetHelper.java#L131-L133

PR #262, opened 2024-02-16
feat: create ref from commit id

### LINKED ISSUE #237: Cannot branch from point in time (specific commit)
Currently, MMS does not support branching from a specific commit, only head of ref.

https://github.com/Open-MBEE/mms/blob/b538d2416766965d848f3edc4b6290302b1aaf10/crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java#L143-L147

borozcod:
### What changed?
Added functionality to be able to make a ref from a commit. 

While developing this feature I identified #261. I've added a test case for checking elements at a specific commit, as well as checking elements made under a branch created from a commit. Both of these tests are currently failing. Fixing #261 should resolve both of these tests.

With this feature, you'll be able to create a ref from a commit like this:
```
POST http://{{host}}/projects/{{projectId}}/refs
Authorization: Basic {{username}} {{password}}
Content-Type: application/json

{
  "refs": [
    {
      "name": "new_branch_from_commit",
      "type": "Branch", // or Tag
      "parentCommitId": "5d374742-52ba-4f01-b225-7af28dfbab33",
      "id": "new_branch_from_commit_id"
    }
  ]
}
```

Fixes #237

### DIFF
--- a/crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
+++ b/crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
@@ -5,11 +5,7 @@ import java.util.Collection;
 import org.openmbee.mms.core.config.Constants;
 import org.openmbee.mms.core.config.ContextHolder;
 import org.openmbee.mms.core.config.Formats;
-import org.openmbee.mms.core.dao.BranchDAO;
-import org.openmbee.mms.core.dao.BranchIndexDAO;
-import org.openmbee.mms.core.dao.CommitDAO;
-import org.openmbee.mms.core.dao.NodeDAO;
-import org.openmbee.mms.core.dao.NodeIndexDAO;
+import org.openmbee.mms.core.dao.*;
 import org.openmbee.mms.core.exceptions.BadRequestException;
 import org.openmbee.mms.core.exceptions.DeletedException;
 import org.openmbee.mms.core.exceptions.InternalErrorException;
@@ -18,6 +14,7 @@ import org.openmbee.mms.core.objects.EventObject;
 import org.openmbee.mms.core.objects.RefsResponse;
 import org.openmbee.mms.core.services.BranchService;
 import org.openmbee.mms.core.services.EventService;
+import org.openmbee.mms.core.services.NodeService;
 import org.openmbee.mms.data.domains.scoped.Branch;
 import org.openmbee.mms.data.domains.scoped.Commit;
 import org.openmbee.mms.data.domains.scoped.Node;
@@ -45,6 +42,15 @@ public class DefaultBranchService implements BranchService {
     private NodeIndexDAO nodeIndex;
 
     protected Collection<EventService> eventPublisher;
+    protected NodeGetHelper nodeGetHelper;
+
+
+
+    @Autowired
+    public void setNodeGetHelper(NodeGetHelper nodeGetHelper) {
+        this.nodeGetHelper = nodeGetHelper;
+    }
+
 
     @Autowired
     public void setBranchRepository(BranchDAO branchRepository) {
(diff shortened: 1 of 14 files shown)


crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
```diff
@@ -145,18 +152,26 @@ public RefJson createBranch(String projectId, RefJson branch) {
...
-            });
+            if(branch.getParentCommitId() != null){
+                Optional<Commit> commitRequestId = commitRepository.findByCommitId(branch.getParentCommitId());
+                commitRequestId.ifPresentOrElse(commit -> {
+                    Optional<Commit> parentCommit = commitRepository.findByRefAndTimestamp(refOption.get(), commit.getTimestamp());
+                    parentCommit.ifPresent(parent -> {
+                        b.setParentCommit(parent.getId());
+                        branch.setParentCommitId(parent.getCommitId());
+                    });
+                },
```
borozcod, line 166: This method was updated so that if a RefJson has a parentCommitId, we don't set the new branch's parentCommitId to the the latest from the parent ref.

dlamoris:
i would suggest leaving the current crud test collection as is and make a new collection just for testing elements at commit, otherwise crud gets increasingly hard to follow on what element is supposed to be where - I think I'll just make a new collection as part of #261, then this can add to that collection by branching from the tests for getElements at commitId

crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
```diff
@@ -181,6 +196,55 @@ public RefJson createBranch(String projectId, RefJson branch) {
...
+        Map<String, ElementJson> nodeCommitData = new HashMap<>();
+        for (ElementJson element : result) {
+            nodeCommitData.put(element.getId(), element);
+        }
+
+        // Update database table to match index
+        for (Node node : nodes) {
+            if(nodeCommitData.containsKey(node.getNodeId())){
+                node.setDocId(nodeCommitData.get(node.getNodeId()).getDocId());
+                node.setDeleted(false);
```
dlamoris, line 245: also set node lastCommitId to be the element json's commitId

dlamoris: need to do what line 189 is doing also - and not do line 189 if it wasn't branch from latest

example/crud.postman_collection.json
```diff
@@ -1276,13 +1252,50 @@
...
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"return latest elements on refa\", function () {",
+							"    var jsonData = pm.response.json();",
+							"    pm.expect(jsonData.commitId).to.eql(pm.environment.get(\"commitXYRef\"))",
+							"",
+							"    pm.environment.set(\"commitXYRefElements\", pm.response.json());",
```
dlamoris: using the entire json here to compare with later won't work since the return elements' _refId should reflect the context they're gotten in - getting the elements from a branch will have slightly different json even if they're the same elements (the _refId will be different)

crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
```diff
@@ -167,11 +182,14 @@ public RefJson createBranch(String projectId, RefJson branch) {
         try {
             branchIndex.update(branch);
             branchRepository.save(b);
-            Set<String> docIds = new HashSet<>();
-            for (Node n: nodeRepository.findAllByDeleted(false)) {
-                docIds.add(n.getDocId());
+            if(branch.getParentCommitId() == null) {
```
dlamoris: the parent commitId would never be null here since it's set at line 172, should add a boolean to keep track above

crud/src/main/java/org/openmbee/mms/crud/services/DefaultBranchService.java
```diff
@@ -181,6 +199,62 @@ public RefJson createBranch(String projectId, RefJson branch) {
...
+                node.setLastCommit(nodeCommitData.get(node.getNodeId()).getCommitId());
+                node.setDeleted(false);
+            } else {
+                node.setDeleted(true);
+            }
+        }
+        nodeRepository.updateAll(nodes);
+
+        Set<String> docIds = new HashSet<>();
+        for (Node n: nodeRepository.findAllByDeleted(false)) {
```
dlamoris: you can add to docIds as you're looping through the nodes so you don't have to call nodeRepository again

borozcod:
Thank you for the feedback @dlamoris. I've gone ahead and committed my changes.

dlamoris (approved):

PR #263, opened 2024-02-17
fix get elements at commit if element was deleted after commit

### LINKED ISSUE #261: Missing check for time of element deletion while getting all elements under commitId

dlamoris:
Checks if element was deleted before or after requested commit if it's currently deleted and last updated time was before commit.
Added new test collection
fixes #261

### DIFF
--- a/crud/src/main/java/org/openmbee/mms/crud/services/NodeGetHelper.java
+++ b/crud/src/main/java/org/openmbee/mms/crud/services/NodeGetHelper.java
@@ -132,9 +132,14 @@ public class NodeGetHelper extends NodeOperation {
                 Optional<ElementJson> e = nodeIndex.getElementLessThanOrEqualTimestamp(nodeId,
                     formatter.format(time), refCommitIds);
                 if (e.isPresent()) { // found version of element at commit time
-                    //TODO determine if element was deleted at the time?
-                    e.get().setRefId(ContextHolder.getContext().getBranchId());
-                    info.getActiveElementMap().put(nodeId, e.get());
+                    Instant realModified = Instant.from(formatter.parse(e.get().getModified()));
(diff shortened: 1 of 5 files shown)


crud/src/main/java/org/openmbee/mms/crud/services/NodeGetHelper.java
```diff
@@ -13,14 +13,24 @@
...
 import org.openmbee.mms.json.ElementJson;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import static org.openmbee.mms.core.config.ContextHolder.getContext;
 
 @Service
 public class NodeGetHelper extends NodeOperation {
 
+    protected CommitIndexDAO commitIndex;
```
HuiJun, line 27: :(

HuiJun (approved):

crud/src/main/java/org/openmbee/mms/crud/services/NodeGetHelper.java
dlamoris: it'll get refactored when we merge in to the boeing branch anyways

PR #274, opened 2024-03-11
Add Jacoco

### LINKED ISSUE #255: add jacoco
can use jacoco agent to instrument the service while postman tests are being run
https://www.eclemma.org/jacoco/trunk/doc/agent.html
the output .exec file can be carried over to the scan job, then use the jacoco cli report command to create report xml after gradle build creates class files
https://www.jacoco.org/jacoco/trunk/doc/cli.html

karicordes:
Added commands to Dockerfile to generate jacoco .exec file, and commands to .circleci/config.yml to generate xml report from .exec file

Fixes #255


.circleci/config.yml
```diff
@@ -43,6 +46,14 @@ jobs:
             docker run --volumes-from mms_test_configs --network container:mms -t postman/newman run artifacts.postman_collection.json -e test-env.json --delay-request 500
             docker run --volumes-from mms_test_configs --network container:mms -t postman/newman run elastic.postman_collection.json -e test-env.json --delay-request 500
             docker run --volumes-from mms_test_configs --network container:mms -t postman/newman run groups.postman_collection.json -e test-env.json --delay-request 500
+            
+            docker stop mms
+            mkdir -p jacocoOutput
+            docker cp mms:/mms/jacocoOutput/jacoco-it.exec ./jacocoOutput/jacoco-it.exec
+            cp -r ./*/build/classes/* TEMPORARYclassfiles/
```
dlamoris: because the class files aren't actually built here, you want to run the cli report in the scan job, after the the run step - instead of checkout in scan, you can attach workspace (like the deploy_snasphot job) 
also you want to get only the build/classes/java/main so it doesn't count test files

dlamoris (changes requested):
also can you add the line 
sonar.coverage.jacoco.xmlReportPaths=jacocoOutput/jacocoXMLReport
in sonar-project.properties, this should let sonarcloud find the report

.circleci/config.yml
dlamoris: add requires: build_and_test to scan in the workflows section so scan comes after build_and_test

karicordes: That makes sense. Thank you!

.circleci/config.yml
```diff
@@ -56,7 +63,13 @@ jobs:
           - checkout
           - run:
                 name: Build classes to scan
-                command: ./gradlew build -x test
+                command: |
+                    ./gradlew build -x test
+                    mkdir -p TEMPORARYclassfiles
+                    cp -r */build/classes/java/main/* TEMPORARYclassfiles/
+                    java -jar jacocoJars/jacococli.jar report jacocoOutput/jacoco-it.exec --classfiles TEMPORARYclassfiles/ --xml ./jacocoOutput/jacocoXMLReport
```
dlamoris, line 71: need to replace -checkout with attach_workspace at line 63 to carry over files from build step (you can see steps that are failing on the conversation page, the link will take you to circleci page that shows more details as it runs)

Dockerfile
```diff
@@ -5,5 +5,6 @@ RUN ./gradlew --no-daemon bootJar --warning-mode all
 
 RUN find . -type f -name example-*.jar -not -iname '*javadoc*' -not -iname '*sources*' -exec cp '{}' '/app.jar' ';'
 ENV JDK_JAVA_OPTIONS "-XX:MaxRAMPercentage=90.0"
-ENTRYPOINT ["java", "--add-opens", "java.base/java.lang=ALL-UNNAMED", "-jar", "/app.jar"]
+COPY jacocoJars/jacocoagent.jar ./jacocoagent.jar
+ENTRYPOINT ["java", "--add-opens", "java.base/java.lang=ALL-UNNAMED", "-javaagent:jacocoJars/jacocoagent.jar=destfile=jacocoOutput/jacoco-it.exec,output=file", "-jar", "/app.jar"]
```
dlamoris, line 8: double check the javaagent location inside the container, the copy step copies it into this directory

dlamoris: i think line 8's copy is actually redundant since line 2 already copies everything

dlamoris:
the sonarcloud context wasn't working with a pr from a fork, i pushed this branch to the main repo with #275 will merge that instead

PR #277, opened 2024-04-17
WIP: Feature/group permissions2.0

Enquier:
Rebasing of groupPermissions adding the following in support of Local (mms-only) Groups:

1. Authentication refactor to support clear User vs Auth workflows
2. Group refactor adding API capabilities
3. Group permission management (i.e. Admin Users who will be able to add/remove members from local (mms-only) groups).


HuiJun:
Why is gradlew an empty file now?

PR #278, opened 2024-04-17
fix commit retrieval loop.

Enquier:
This loop was happening in the background where commits were just piling up or never returning unless a limit was set.


rdb/src/main/java/org/openmbee/mms/rdb/repositories/commit/CommitDAOImpl.java
```diff
@@ -156,7 +157,7 @@ public List<Commit> findByRefAndTimestampAndLimit(Branch ref, Instant timestamp,
             currentRef = ref.getParentRefId();
             currentCid = ref.getParentCommit();
 
-            if (currentRef == null) {
+            if (currentRef == null || currentRef.equals(Constants.MASTER_BRANCH)) {
```
dlamoris: This breaks the tests since it stops at a branch whose parent is master.
master branch is created with parent ref being null, it may be you had some bad db data from a dev branch for some reason had master with master parent

dlamoris (approved):

